PR #151, opened 2015-05-30
Initial modifications to support py2neo 2.x

bleib1dj:
## Overview ##
This PR provides the updates necessary to move towards py2neo 2.x. Functionality remains largely backwards compatible except for the known move from raising a `CypherException` to raising an `InvalidSyntax` when a cypher query has invalid syntax. This provides a more clear indication as to what is going wrong.
 
## Unresolved Issues ##
* Pypy tests are not passing on TravisCI
* Need to transition `START`queries to `MATCH`
* Documentation surrounding Neo4j 2.2.x requirements
  * Information primarily directing users to py2neo's docs on setting a new password for a new db

I believe 2 and 3 in the list above are outside the scope of this PR and should be moved into separate issues. As for pypy support I'm not sure what is causing the issue and am not up on pypy so I was hoping during the review someone could assist in pointing out what needs to be changed to fix the failing tests.

### DIFF
--- a/neomodel/contrib/semi_structured.py
+++ b/neomodel/contrib/semi_structured.py
@@ -52,17 +52,17 @@ class SemiStructuredNode(StructuredNode):
     def inflate(cls, node):
         props = {}
         for key, prop in cls.defined_properties(aliases=False, rels=False).items():
-            if key in node._properties:
-                props[key] = prop.inflate(node._properties[key])
+            if key in node.properties:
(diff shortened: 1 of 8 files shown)


robinedwards:
Thank you so much for all your hard work this is amazing effort, I've merged and remove pypy from travis for the time being

bleib1dj:
Not a problem and cool :)

PR #164, opened 2015-06-17
README updates based on Neo4j 2.2 and py2neo 2 support

bleib1dj:


README.rst
```diff
@@ -43,6 +43,53 @@ To install from github::
...
+Authentication
+--------------
+Please note that if you are utilizing Neo4j version 2.2 or newer there are
+some additional setup steps necessary relating to authentication. As of version 2.2
+Neo4j authentication is activated by default on new instances. Please follow the
+outstanding documentation provided by py2neo's Authentication_
+section to setup new credentials. If you are utilizing a hosted service this
+is most likely already taken care for you.
+
+.. _Authentication: http://py2neo.org/2.0/essentials.html#authentication
```
robinedwards: I am just wondering if we need to cover how to set the username and password from neomodel in the getting started section: https://github.com/robinedwards/neomodel/blob/master/doc/source/getting_started.rst

README.rst
```diff
@@ -43,6 +43,53 @@ To install from github::
...
+a few more capabilities on accessing the results. The draw back is
+that now accessing the above query at ``res[0][0]`` will result in an ``IndexError``
+due to the new response being an empty ``RecordList``. Py2neo has however provided
+a solution to this which is the one_ method. This method can be called on the
+cypher query response to achieve the old ``res[0][0]`` functionality. So in
+version 2 the above query would look like::
+
+    query = 'MATCH (a:Profile {username: "%s"}) RETURN a' % (username)
+    res, col = db.cypher_query(query)
+    end_result = res.one()
```
robinedwards: Not a note on the docs but more the implementation. Do we want to expose these py2neo objects directly from the neomodel API or do we want to do decouple them? I've always considered different backends or even embedding our own client inside neomodel eventually.

bleib1dj:
That's a good question. Decoupling may result in some redundant work but could be beneficial from a developers perspective where they only have to worry about the docs from one package. This will grow though from just these few items to also include the exceptions that are thrown by py2neo which are useful in debugging since they give more specific information. 

As for different backends are there any that exist today outside of py2neo, even those not neo4j based, that are under active development? It may be nice to build an embedded client but I think there are other portions of the package that we could focus on improving initially that would have a larger impact.

Another opportunity may be to align our development and community efforts more heavily with py2neo. Since both projects have relatively small communities but both seem to be the leading systems relating to Neo4j it'd be nice to integrate more tightly to cut down on redundancy, improve performance, coalesce around best practices, streamline documentation, enable better support for python frameworks (django, flask, ect), and grow python neo4j support together. Then once we've established a strong community around neo4j support we could use the definitions we've established together to branch out.

But even with that approach I would agree that some level of decoupling would be good so that down the line additional backends could be developed and dropped in that allowed for alternatives to neo4j. I'm not up on how pervasive Cypher is throughout the graph community though and we should be aware that without a common language like SQL it may be difficult to support multiple graph backends if they all require different language integrations.

README.rst
bleib1dj: I'd be happy to move and/or add it there.

bleib1dj:
Just found these projects starting up too, which we should keep an eye on and maybe get some info from nigelsmall on [neo4j-python-driver](https://github.com/neo4j/neo4j-python-driver) [neo4j-ogm](https://github.com/neo4j/neo4j-ogm)

robinedwards:
Yea by backend I meant client as opposed to actual database. This one looked promising also: https://github.com/jakewins/neo4jdb-python.

I am happy to keep it coupled to py2neo for now its just a point that I thought maybe worth raising.

bleib1dj:
Yeah it's a good point that should have some further discussion especially with neo4j appearing to transition to neo4j authorized/supported drivers. Figuring out what path to take with that will also help to plan out what features should be focused on and what the strategy/focus for the project is moving forward. Not sure what the best forum for that will be, ember's community has a pretty interesting process with RFCs https://github.com/emberjs/rfcs but it may be overkill and a google group may suffice.

PR #169, opened 2015-07-19
rsplit on @ char, this allows passing passwords with "@" chars

Hexadite-Max:
otherwise the split is done in the wrong place, cutting the password itself


bleib1dj:
@Hexadite-Max awesome and nice catch. Thanks for making the PR.

PR #170, opened 2015-07-19
on _count() call make sure "order_by" is not built into the query

Hexadite-Max:
passing order_by results in an invalid cypher count query


bleib1dj:
Awesome, thanks Max!

PR #177, opened 2015-08-09
check "order_by" on source and not self

Hexadite-Max:
when the NodeSet is a "complex" one built on top of a Traversal etc. this call fails as the check is on the self.node_set, when the source is an inner NodeSet.


bleib1dj:
Awesome, tahnks @Hexadite-Max

PR #178, opened 2015-08-09
feature wildcard and non explicit traversal

Hexadite-Max:
add support for additional relationship types:
* ``None`` = [unnamed directional relationships](http://neo4j.com/docs/stable/query-match.html#match-outgoing-relationships) by target node type
* ``*`` = [variable_length](http://neo4j.com/docs/stable/introduction-pattern.html#_variable_length) by target node type

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -5,16 +5,34 @@ import inspect
 OUTGOING, INCOMING, EITHER = 1, -1, 0
 
 
-def rel_helper(**rel):
-    if rel['direction'] == OUTGOING:
+def rel_helper(lhs, rhs, ident=None, relation_type=None, direction=None, **kwargs):
+    """
+    Generate a relationship matching string, with specified parameters.
+    Examples:
+    relation_direction = OUTGOING: (lhs)-[relation_ident:relation_type]->(rhs)
+    relation_direction = INCOMING: (lhs)<-[relation_ident:relation_type]-(rhs)
+    relation_direction = EITHER: (lhs)-[relation_ident:relation_type]-(rhs)
+
+    :param lhs: The left hand statement.
+    :type lhs: str
+    :param rhs: The right hand statement.
+    :type rhs: str
+    :param ident: A specific identity to name the relationship, or None.
+    :type ident: str
+    :param relation_type: None for all direct rels, * for all of any length, or a name of an explicit rel.
+    :type relation_type: str
+    :param direction: None or EITHER for all OUTGOING,INCOMING,EITHER. Otherwise OUTGOING or INCOMING.
+    :rtype: str
+    """
+
+    if direction == OUTGOING:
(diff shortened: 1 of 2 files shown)


neomodel/match.py
```diff
@@ -7,13 +7,24 @@
...
-        stmt = '-[{0}:{1}]->'
+        stmt = '-{0}->'
     elif rel['direction'] == INCOMING:
-        stmt = '<-[{0}:{1}]-'
+        stmt = '<-{0}-'
     else:
-        stmt = '-[{0}:{1}]-'
-    ident = rel['ident'] if 'ident' in rel else ''
-    stmt = stmt.format(ident, rel['relation_type'])
+        stmt = '-{0}-'
```
bleib1dj: Could this just be simplified to always utilize the bidirectional statement of `'-{0}-'`? I believe the [performance is the same regardless of direction](http://graphaware.com/neo4j/2013/10/11/neo4j-bidirectional-relationships.html). Unless this is to help simplify the logic of the helper to a user. @Hexadite-Max What are your thoughts?

Hexadite-Max: I suppose direction should be explicit, defaulting to ```(x)--(y)``` may have unexpected results for users, if I understand your question correctly.

bleib1dj: I think defaulting to `(x)--(y)` is okay since there isn't a performance issue and the user will still be able to traverse in either direction they'd like. I guess I'm asking if we need to define it as 

    if rel['direction'] == OUTGOING:
        stmt = '-{0}->'
    elif rel['direction'] == INCOMING:
        stmt = '<-{0}-'
    else:
        stmt = '-{0}-'

instead of removing the need for OUTGOING and INGOING and simply define it as

    stmt = '-{0}-'

Since neo4j doesn't seem to care internally about direction. Unless I'm reading 

```
Relationships in Neo4j can be traversed in both directions with the same speed. Moreover, 
direction can be completely ignored. Therefore, there is no need to create two different 
relationships between nodes, if one implies the other.
```
incorrectly. Do you know if using `(x)--(y)` causes unexpected behavior when used in place of `(x)<--(y)' or `(x)-->(y)`?

Hexadite-Max: I think there is no penalty, but there is a logical difference implied by a relationship direction. i.e. if the relationships were defined having a direction, the simple truth is that ```(x)--(y)``` and ```(x)-->(y)``` will have different results, so changing this would be breaking. it is possible however to do `rel.get('direction')` so it is not mandatory: defaulting to `(x)--(y)`

P.S. how about changing the func definition to `def rel_helper(lhs, rhs, ident=None, relation_type=None, direction=None)` so mandatory and optiona params are explicit?

bleib1dj: Gotcha and I'm a fan of that change.

PR #181, opened 2015-08-20
Django settings

midnightlynx:
Added optional import of NEO4J_REST_URL through django.conf.settings.
If django is not installed and properly configured, or if NEO4J_REST_URL isn't defined in settings, it falls back to getting this value from the environment variable as before.


nikbora:
Hey guys, I am curious why there is a need to add django dependencies to neomodel? Isn't neomodel meant to be standalone (e.g. we use it heavily from flask). 

Perhaps it would be better to have a django-neomodel plugin of some sort to deal with django specifics, a bit like you can get flask-sqlalchemy for sqlalchemy?

robinedwards:
I agree with @nikbora, however this PR doesn't add it as a dependency as such, it would be nice to see specific modules for integration into the frameworks. Perhaps we should create a neomodel organisation and then flask_neomodel and django_neomodel can live under there?

nikbora:
:+1: I agree this particular PR isn't too bad and is very guarded, but can see it being a precedent for more djangoey stuff in neomodel core hence the question :smile:

midnightlynx:
I'm cool with helping with an integration module to do this instead. :) Maybe the support bits for django signals can be shifted there eventually as well.

bleib1dj:
I concur. @robinedwards do you want to create the org?

robinedwards:
@bleib1dj I've created the organisation, sorry for the delay away on vacation

bleib1dj:
@robinedwards not a problem :). Thanks for setting it up!

midnightlynx:
How do I find the org?

bleib1dj:
https://github.com/neomodel

PR #184, opened 2015-09-02
performance enhancements for bulk operations 

Hexadite-Max:
**parameterized bulk queries**
utilize passing property maps for MERGE and CREATE queries
see: http://neo4j.com/docs/stable/cypher-parameters.html#_create_multiple_nodes_with_properties
and: http://neo4j.com/docs/stable/query-set.html#set-adding-properties-from-maps
this provides significant speed up for hundreds of nodes

**add "lazy" ID only query results**
significant (seconds) speed up for hundreds of nodes where property values are not required for further operations

**cache all most used `defined_properties()`**
this method is called multiple times for each node class `init()`, this provides additional optimization when inflating nodes.

### DIFF
--- a/neomodel/contrib/semi_structured.py
+++ b/neomodel/contrib/semi_structured.py
@@ -50,27 +50,33 @@ class SemiStructuredNode(StructuredNode):
 
     @classmethod
     def inflate(cls, node):
-        props = {}
-        for key, prop in cls.defined_properties(aliases=False, rels=False).items():
-            if key in node.properties:
-                props[key] = prop.inflate(node.properties[key])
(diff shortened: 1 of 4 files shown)


smcclstocks:
All of this looks super useful for my needs as well. I was about to sub-class different neomodel classes to pull it off but i'd much prefer it native. Any chance we can get this included & cut a release? Thanks!

bleib1dj:
@smcclstocks definitely I've been tied up the last few weeks but have been meaning to get a new release out with all of the new additions. I'll hopefully have a chance to get the new version out this weekend.

PR #186, opened 2015-09-16
Add "db_property" as parameter to Property.

fpieper:
You are now able to use different names for the property in the database and the python node attribute.
	This is quite useful, for "hiding" the property behind a "python property" subsequently. So you don't need to rename you database fields.
	For example renaming the property "name" to "name_" and adding a "python @property" "name" afterwards.
	Usages:
		name_ = StringProperty(db_property="name")
		name = StringProperty(db_property="full_name")

### DIFF
--- a/doc/source/properties.rst
+++ b/doc/source/properties.rst
@@ -57,3 +57,21 @@ Allows aliasing to other properties can be useful to provide 'magic' behaviour, 
         name = AliasProperty(to='full_name')
 
     Person.nodes.filter(name='Jim') # just works
+    
+Independent database property name
+================================
+
+You can specify an independent property name with 'db_property', which is used on database level. It behaves like Django's 'db_column'.
+This is useful for e.g. hiding graph properties behind a python property::
+
+    class Person(StructuredNode):
+        name_ = StringProperty(db_property='name')
+        
+        @property
+        def name(self):
+            return self.name_.lower() if self.name_ else None
+
+        @name.setter
+        def name(self, value):
+            self.name_ = value
+
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -145,12 +145,12 @@ class StructuredNode(NodeBase):
         # create or update instance node
         if hasattr(self, '_id'):
             # update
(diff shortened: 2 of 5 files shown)


fpieper:
I thought about the following naming options:
db_property_name (too long)
property_name (missing hint to database)
db_property (I selected that)
db_name (it's confusing about database name)

What do you think?

robinedwards:
Interesting there is also the AliasProperty but I see this solves the problem in a different way

fpieper:
The idea is to remove the direct coupling between the property name in the Python code and the property name in the actual database.

So you start with the following definition:
```
name = StringProperty()
```

And after you are in production and have a lot of data in your database you need to do the following:

```
name_ = StringProperty(db_property="name")
@property
def name(self):
    return self.name_.lower()
```

Without "db_property" you would need to change all names in the database.
That is not possible with AliasProperty.

robinedwards:
Yes good point its identical to django's db_column

https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.Field.db_column

Would you be able to include some documentation for it? I think its a nice addition what do you think @bleib1dj?

fpieper:
Yes, I can add some documentation. Do you want to change the name of the parameter? (like written above).

I knew the "db_column" of django and searched for a way to use that with neomodels ;)

robinedwards:
No I think db_property makes more sense are there are no columns in a graph database

fpieper:
Sorry that was a little bit confusing. I was referencing to my previous post.

I mean whether you would prefer "property_name" over "db_property".

robinedwards:
I think db_property is better :-)

bleib1dj:
Sounds like a good addition to me.

fpieper:
I added a chapter to the documentation of properties

robinedwards:
Super could you add a quick test case for this?

fpieper:
Yes, I added a test case with low level database checking.

robinedwards:
Awesome cheers for that! Sorry for the delay

PR #196, opened 2015-11-07
Email and Regex properties

rafaelpivato:
Added some extra properties:
- `NormalProperty`
- `RegexProperty`
- `EmailProperty`

They are all based on `NormalProperty`, which serves as a base class for properties using the same method for either `inflate` or `deflate`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ development.env
 .idea
 .python-version
 .ropeproject
+\#*\#
--- a/doc/source/properties.rst
+++ b/doc/source/properties.rst
@@ -20,6 +20,10 @@ You may provide arguments using a wrapper function or lambda::
 
         my_datetime = DateTimeProperty(default=lambda: datetime.now(pytz.utc))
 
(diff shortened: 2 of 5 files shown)


rafaelpivato:
Forgot to mention that we have `default_now` also for `DateTimeProperty` class.

robinedwards:
Super thanks @rafaelpivato.

Some thoughts on this, not sure what the default_now offers that can't be achieved with datetime.now, the current method in neomodel follows django's convention which seems a bit more intuitive.

Regex property looks very useful though, I am a bit undecided if something such as an email field should live in this framework or a form library for example.

@bleib1dj what are your thoughts? As for normalproperty I think I'd like to keep a single common base class

shayneoneill:
The normalize abstraction is excellent as a simple platform to build more complex types, as is the regex type

robinedwards:
Thanks I've merged this in manually, sorry for the delay. If your still using neomodel and wish to help out please get in touch :-)

Issue #214, opened 2016-07-22
Random ordering


romaintha:
Hi,
I would like to implement the random ordering for neomodel, similar to Django order_by("?"). I found this regarding Cypher :
http://jexp.de/blog/2014/03/random-thoughts-ordering-in-cypher/
I have a very limited insight of neomodel, but it does not seem easy to implement. Any suggestions where to start?

romaintha:
Hi @the4thv ,
I do agree that the generated load might become a problem. But I guess it would be great to have the possibility to do it, and the performance consideration would be let to anyone using it.
I envisioned the solution you proposed. But I was thinking to mimic as much as possible the django order_by("?").  As far as I see the NodeSet class of neomodel, using order_by(prop) checks that prop is indeed a property.  So there would be a need to rework that. In addition looking at the QueryBuilder class, and how the order_by query is built, it seems complicated to implement here the method I attached previously. 
What do you think?

the4thv:
@romaintha , yes. I'm actually working on neomodel for Neo4j 3.0 with Robin's blessing and I'd be happy to implement your proposal in that. I'm reworking the basic logic to bring it more inline with new Neo4j cypher, including the QueryBuilder. It won't be too hard for me to take into consideration adding support for this.

romaintha:
@the4thv then I ll work on that and let you know when I have a proposal

the4thv:
@romaintha cool - thanks :)

the4thv:
Thank you for your proposal. I'll look to implement that in the new alpha version of neomodel and we can review how it works with Neo4j.

robinedwards:
Merged thanks ever so much :-)

PR #215, opened 2016-07-23
allow for random ordering, mimicing the Django order_by('?')

romaintha:
Following #214 , here is a proposal to handle random ordering mimicing the Django order_by('?')


robinedwards:
Thanks dude, much appreciated

PR #217, opened 2016-09-18
In case of numeric properties default && required check doesn't work as expected

Vorren:
`default = 0` evaluates to false, which it is not when we're talking about numbers.


robinedwards:
Great spot thank you!

PR #221, opened 2016-12-15
Bugfix empty filter append

siddharthm:
Don't append empty filters to filters list. Otherwise this leads to self.filters not being evaluated as false leading to empty where statements in cypher query


robinedwards:
Brilliant thank you!

PR #225, opened 2016-12-19
python3: use .items() instead of deprecated .iteritems()

eamonnmag:
* Makes the library python 3 compatible, without this change, the error 
  ```Error “ 'dict' object has no attribute 'iteritems' ”``` when trying to add properties
  to a relation.


robinedwards:
Super thanks

PR #243, opened 2017-02-28
added UniqueIdProperty

gmauro:
I got an exception "neomodel.exception.ConstraintValidationFailed" trying to do Batch nodes operations with create_or_update and UniqueIdProperty.
To reproduce it, I have added the UniqueIdProperty to the test file.

Any hint?

### DIFF
--- a/test/test_batch.py
+++ b/test/test_batch.py
@@ -1,8 +1,9 @@
-from neomodel import (StructuredNode, StringProperty, IntegerProperty)
+from neomodel import (StructuredNode, StringProperty, IntegerProperty, UniqueIdProperty)
 from neomodel.exception import UniqueProperty, DeflateError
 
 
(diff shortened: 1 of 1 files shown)


robinedwards:
Thanks i will look into it

robinedwards:
Hi,

So when you are using the unique id property with a create_or_update you need to specify the uid. Its a merge operation in cypher underhood. In this particular test the email is serving as the unique identifier and as the uid isnt supplied so neomodel generates one.

In a real life situation you would provide the uids you did know and neomodel would generate missing ones. Does that make sense?

I've updated the docs in master added test using uniqueidproperty. Let me know if you are still having issues.

robinedwards:
Will let you know when i have pushed the updated tests and docs

PR #255, opened 2017-06-01
Remove START and node() from cypher query build

the4thv:
Neo4j v3.x does not support either START or node(<id>) therefore breaking relationship traversals, (e.g. node1.test_rel.all()) raising the exception "Internal error - should have used fall back to execute query, but something went horribly wrong" within Neo4j itself.

This fix should be like for like functionality, replace the START part of the query with the format "MATCH (<nodeAlias>) WHERE id(<nodeAlias>)=<id> WITH <nodeAlias>".

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -362,8 +362,17 @@ class QueryBuilder(object):
         query = ''
 
         if 'start' in self._ast:
-            query += 'START '
-            query += ', '.join(self._ast['start'])
+            initial_match = {'match': [], 'where': [], 'with': []}
+
+            for i in self._ast['start']:
+                r = i.split(' = ')
+                initial_match['match'].append('(%s)' % r[0])
+                initial_match['where'].append('id(%s)=%s' % (r[0], r[1].replace('node(','').replace(')','')))
+                initial_match['with'].append(r[0])
+
+            query += 'MATCH %s WHERE %s WITH %s' % (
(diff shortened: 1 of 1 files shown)


the4thv:
Here's an example where it fails. By the way, I'm not using 'replace', I've just replaced the part of match.py which built up a START using node ID.

The patch I've proposed is only 'like for like' solution, where the part of the cypher query you build up the 'START alias=node(id)' which is no longer valid now looks like this:

'MATCH (alias) WHERE id(alias)=node_id WITH alias'

This means the statement in AST should behave identically to before.

[rel_all_example.zip](https://github.com/robinedwards/neomodel/files/1044622/rel_all_example.zip)

the4thv:
If you'd like I could look at the code that build the 'start' element of _ast and turn it from a string into a list of dicts, i.e. [{'alias':'id'}]

robinedwards:
Hey just checked this locally it appears neo4j 3.2 has removed a bunch of deprecated stuff. I am investigating.

the4thv:
Let me know if you would like me to look at any parts of that.

Sent from my iPhone

On 9 Jun 2017, at 19:43, Robin Edwards <notifications@github.com<mailto:notifications@github.com>> wrote:


Hey just checked this locally it appears neo4j 3.2 has removed a bunch of deprecated stuff. I am investigating.

robinedwards:
Right used your with work around thanks for that. Its not ideal but it will do for now. Cheers!

PR #261, opened 2017-07-26
Test for model is None in Traversal.match()

pvanheus:
Since `RelationshipTo` and `RelationshipFrom` both set model to None in their constructors, the test needs to be for `model is None` not just for the `'model'` key being missing.

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -652,7 +652,7 @@ class Traversal(BaseSet):
         :param kwargs: see `NodeSet.filter()` for syntax
         :return: self
         """
-        if 'model' not in self.definition:
+        if 'model' not in self.definition or self.definition['model'] is None:
             raise ValueError("match() only available on relationships with a model")
         if kwargs:
             output = process_filter_args(self.definition['model'], kwargs) 
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -652,9 +652,9 @@ class Traversal(BaseSet):
         :param kwargs: see `NodeSet.filter()` for syntax
         :return: self
         """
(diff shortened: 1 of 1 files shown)


robinedwards:
Thanks, this appears to brake a test would you mind fixing it?

pvanheus:
Thanks @robinedwards, the problem is that the test in question calls `.match()` on a relationship that has no associated model. Looking into fixing this there are two options:

1. `match()` with no argument means return the nodes yielded by traversing the relationship, so the test is correct and the `match()` code needs to only throw an error if `kwargs` is defined and `self.definition['model']` does not exist or is `None`.

2. `match()` with no arguments is an invalid call, so the test needs to be updated to not call this.

Which option do you think is correct?

robinedwards:
Thanks @pvanheus apologies was half asleep yesterday.

To me there is no harm in allowing a match with no arguments if they have no rel_model setup... but they should really just use the Traversal() object for that rel. Perhaps I will add a deprecation warning to that at some point but for now I think I have to maintain backwards compatibility.

One thing this has highlighted is that match() is a really bad name, I think filter_by_rel() or something would be more appropriate? Considering deprecating the method and creating another better named one for the 4.0.0 release. What do you think?

Thanks again for this patch

pvanheus:
Well at the moment `match()` seems to do two things:

1. something like `all()` but returning a `NodeSet` like object rather than a list.
2. traverse a relationship using a filter (if you give it a filter expression).

So should `all()` take over case #1 and then `filter()` or `filter_by_rel()` take over case #2?

robinedwards:
Yes exactly

robinedwards:
Will need to add some kind of deprecation warning for this

PR #273, opened 2017-10-08
Add feature for removing constraints and indexes from schema

laky55555:
This feature is supposed to add functionality similar to neomodel_install_labels, just instead of creating new indexes and constraints it removes all existing ones.
Code is modeled after install_labels example.
Test is also added that remove all indexes and constraints and then reinstall all removed again not to mess up with other tests.
This feature was missing because manually removing installed labels especially when there is a lot of them is tedious work.

### DIFF
--- a/doc/source/getting_started.rst
+++ b/doc/source/getting_started.rst
@@ -67,6 +67,14 @@ Neomodel provides a script to automate this::
 
 It is important to execute this after altering the schema. Keep an eye on the number of classes it detects each time.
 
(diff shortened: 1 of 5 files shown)


robinedwards:
Thanks for this, any chance you could add some documentation to the docs/ so people know about the feature.

laky55555:
I added short doc into getting_started.rst.
Feel free to move it to some other place and modify something.

robinedwards:
Super thank you :-)

PR #288, opened 2017-12-10
Several updates, mainly testing-related

funkyfuture:
ciao, i just wanted to check if i can expect neomodel to work with neo4j 3.3 and ended up with this. it's a bunch of patches that should be self-explanitory and cherry-pickable.


robinedwards:
This is fantastic, thanks ever so much for your work.

Issue #283, opened 2017-11-13
Inheritance breaks


aanastasiou:
As far as I can tell, neomodel is not handling inheritance as expected (or intended?) and for large models this is becoming a problem very quickly.

Specifically, if i have a constraint like:

    class AssociableNode(neomodel.StructuredNode):
        __abstract_node__ = True
        [. . .]

    class Content(AssociableNode):
       [. . . ]

    class ImageContent(Content):
        [. . .]

    class TextContent(Content):
       [. . .]

    class MyNode(neomodel.StructuredNode):
        content = neomodel.Relationship("Content", "HAS_CONTENT")

Then, everything seems alright server side, but when you try to retrieve `MyNode.content[0]` you only get the attributed of "Content", NOT whatever CAN BE at the end of that relationship. TextContent and ImageContent are subclasses of Content and therefore are perfectly valid as contents on My Node.

I have tried various alternatives of expressing the same thing, including making the associable a mixin and different combinations of which classes are considered `__abstract_class__` but the end result is the same: Once I try to access `MyNode.content[0]` I get `AttributeError: type object [. . .] has no attribute '__label__'`.

Any ideas?

aanastasiou:
Having looked a little bit more into the code, I can see how implementing this functionality is a major piece of work. `inflate` constructs an object based on the `cls` declaration. 'Normal' inheritance is easy because it retraces in a straight line from a given class **upwards** in the hierarchy. But, enforcing the constraint of "cls and any other class that **descends from** cls" is looking across the hierarchy. 

The solution likely involves a change to the way labels are applied in the database to be able to infer the referring class from the label (or combination of labels?) and then inflate a copy of **that specific** class (not the abstract one).

I cannot see any document outlining the internals so I thought I might ask, is there such a description available from somewhere? I am trying to understand the bigger picture of how neomodel works by using pyreverse and pycallgraph and that is not at all fun. It is very difficult to contribute without understanding the bigger picture. How are nodes represented? How are relationships represented? What happens upon CRUD for each?

I would be more inclined to contribute to neomodel rather than switch the GOM as this is closest to what I am after in terms of data validation / way of accessing the backend. However, this is already eating a large chunk of my time and I do not think I can apply these changes on my own, any chance for any help with this?

aanastasiou:
A little bit more information about this, as I am trying to make sense of what is going on and potentially try to improve the situation:

[...]

PR #302, opened 2018-02-23
Initial work on issue 283 - "Inheritance Breaks"

aanastasiou:
# Changes
Initial work on rectifying issue 283, outlined [here](https://github.com/neo4j-contrib/neomodel/issues/283) and in more detail [here](https://github.com/aanastasiou/neomodelInheritanceTest).

A "class registry" has been added in `util.Database`. It is populated automatically by `core.NodeMeta.__new__` and subsequently used by `util.Database.cypher_objectaware_query` to properly instantiate query results to objects in a hierarchy.

This is **effected** by `match.QueryBuilder._execute` (which now uses `cypher_objectaware_query`) when the entity is instantiated **from** the database and **enforced** by `relationship_manager.RelationshipManager._check_node` when a connection between two entities is attempted to be established. The latter now tests for `issubclass` rather than `isinstance`.

To keep this functionality consistent, `relationship.StructuredRel.start_node` and `relationship.StructuredRel.end_node` had to also be modified so that they instantiate nodes of the correct type as inferred by the labels retrieved from the database (as opposed to simply looking at the type set when defining the model).

Entity specific `inflate` functions (For StructuredNode and StructuredRel) are still valid in case a node of specific class that is known at run-time needs to be instantiated.

# Open issues

1. The class registry is populated automatically, when classes are "created" at runtime. Consequently, the registry would be unaware of classes whose definition is perfectly alright **but have not been loaded yet** (i.e. not yet `import`ed).
    * This will raise `KeyError` on the class registry anyway but perhaps a more informative neomodel-specific exception is required here (?)
    * If this could be regarded as "expectable behaviour", it would be best if it is mentioned in the documentation

2. There is a corner case where a query might include labels that are associated with an entity in the database but not fully specified in the model (yet). This is similar to the first issue but also covers the case of "additional labels found" (as opposed to just "no matching labels found").
    * Should this raise an exception?
    * Does this warrant a "switch" parameter to turn on a "silent fail"? In this case, a list of results is returned that **could** contain plain simple `Node` objects for those results that could not be resolved.
    * Could as well appear in the documentation as a note or a warning.

### DIFF
--- a/doc/source/relationships.rst
+++ b/doc/source/relationships.rst
@@ -99,7 +99,7 @@ further defined. ::
         Base class for defining some basic sort of an actor in a system.
         
         The base actor is defined by its name and a `friends_with` 
-        relationship
+        relationship.
         """
         name = neomodel.StringProperty(required = True, unique_index = True)
         friends_with = neomodel.RelationshipTo("BasePerson", "FRIENDS_WITH", model = PersonalRelationship)
@@ -107,14 +107,14 @@ further defined. ::
     class TechnicalPerson(BasePerson):
         """
         A Technical person specialises BasePerson by adding their 
-        expertise
+        expertise.
         """
         expertise = neomodel.StringProperty(required = True)
         
     class PilotPerson(BasePerson):
         """
         A pilot person specialises BasePerson by adding the type of 
-        airplane they can operate
+        airplane they can operate.
         """
         airplane = neomodel.StringProperty(required = True)
         
@@ -155,7 +155,7 @@ Automatic class resolution
 --------------------------
 
 Neomodel is able to transform nodes to objects, automatically, via a *node-class registry* 
-that is progreessively built up during the definition of the models.
+that is progressively built up during the definition of the models.
 
 The dictionary provides a mapping from the set of labels associated with a node to the class 
 that is implied by this set of labels. In the example above, the *node-class registry* would contain 
--- a/doc/source/relationships.rst
+++ b/doc/source/relationships.rst
@@ -2,7 +2,8 @@
 Relationships
 =============
 
-Directionless relationships, first argument the class second the neo4j relationship::
+Establishing an undirected relationship between two entities is done via the `Relationship` 
+class. This requires the class of the connected entity as well as the type of the relationship.
 
     class Person(StructuredNode):
         friends = Relationship('Person', 'FRIEND')
@@ -16,8 +17,8 @@ This avoids cyclic imports::
 
 Cardinality
 ===========
-It's possible to (softly) enforce cardinality restrictions on your relationships.
-Remember this needs to be declared on both sides of the definition::
+It is possible to (softly) enforce cardinality constraints on your relationships.
+Remember this needs to be declared on both sides of the relationship definition::
 
     class Person(StructuredNode):
         car = RelationshipTo('Car', 'OWNS', cardinality=One)
@@ -25,16 +26,17 @@ Remember this needs to be declared on both sides of the definition::
     class Car(StructuredNode):
         owner = RelationshipFrom('Person', 'OWNS', cardinality=One)
 
-The following cardinality classes are available:
+The following cardinality constraints are available:
 
 ===================================================  ========================================
 :class:`~neomodel.cardinality.ZeroOrOne`             :class:`~neomodel.cardinality.One`
 :class:`~neomodel.cardinality.ZeroOrMore` (default)  :class:`~neomodel.cardinality.OneOrMore`
 ===================================================  ========================================
 
-If cardinality is broken by existing data a :class:`~neomodel.exception.CardinalityViolation`
+If a cardinality constrain is violated by existing data a :class:`~neomodel.exception.CardinalityViolation`
 exception is raised.
-On attempting to break a cardinality restriction a
+
+On attempting to violate a cardinality constrain a 
 :class:`~neomodel.exception.AttemptedCardinalityViolation` is raised.
 
 Properties
@@ -56,7 +58,8 @@ Neomodel uses :mod:`~neomodel.relationship` models to define the properties stor
     rel.since # datetime object
 
 
-These can be passed in when calling the connect method::
+The data to populate these properties when establishing a connection can be supplied 
+to the `connect` method::
 
     rel = jim.friends.connect(bob,
                               {'since': yesterday, 'met': 'Paris'})
@@ -75,8 +78,8 @@ This is only available for relationships with a defined relationship model::
 Relationship Uniqueness
 =======================
 
-By default in neomodel there is only one relationship of one type between two nodes
-unless you define different properties when calling connect. neomodel utilises `CREATE UNIQUE` in cypher to achieve this.
+By default neomodel applies only one relationship instance between two node instances and 
+this is achieved via use of `CREATE UNIQUE`.
 
 Relationships and Inheritance
 =============================
@@ -179,9 +182,6 @@ If the exception is not handled, it produces an error message that returns the l
 was retrieved from the database as well as the current *node-class registry*. These two pieces of 
 information can be used to debug the model mismatch further.
     
-
-
-        
 
 Explicit Traversal
 ==================
--- a/doc/source/relationships.rst
+++ b/doc/source/relationships.rst
@@ -78,6 +78,111 @@ Relationship Uniqueness
 By default in neomodel there is only one relationship of one type between two nodes
 unless you define different properties when calling connect. neomodel utilises `CREATE UNIQUE` in cypher to achieve this.
 
+Relationships and Inheritance
+=============================
+Neomodel is capable of understanding and resolving inheritance relationships properly.
+
+The following model establishes a BasePerson that can be `friends_with` any class derived 
+from BasePerson. Two concrete classes of BasePerson (TechnicalPerson and PilotPerson) are 
+further defined. ::
+
+
+    class PersonalRelationship(neomodel.StructuredRel):
+        """
+        A very simple relationship between two BasePersons that simply 
+        records the date at which an acquaintance was established.
+        """
+        on_date = neomodel.DateProperty(default_now = True)
+        
+    class BasePerson(neomodel.StructuredNode):
+        """
+        Base class for defining some basic sort of an actor in a system.
+        
+        The base actor is defined by its name and a `friends_with` 
+        relationship
+        """
+        name = neomodel.StringProperty(required = True, unique_index = True)
+        friends_with = neomodel.RelationshipTo("BasePerson", "FRIENDS_WITH", model = PersonalRelationship)
+        
+    class TechnicalPerson(BasePerson):
+        """
+        A Technical person specialises BasePerson by adding their 
+        expertise
+        """
+        expertise = neomodel.StringProperty(required = True)
+        
+    class PilotPerson(BasePerson):
+        """
+        A pilot person specialises BasePerson by adding the type of 
+        airplane they can operate
+        """
+        airplane = neomodel.StringProperty(required = True)
+        
+This means that either of these concrete objects can appear at the end 
+of a ``friends_with`` relationship and be instantiated to the right object.
+
+Here is a minimal example to demonstrate that::
+
+    # Create some technical persons
+    A = TechnicalPerson(name = "Grumpy", expertise = "Grumpiness").save()
+    B = TechnicalPerson(name = "Happy", expertise = "Unicorns"}).save()
+    C = TechnicalPerson(name = "Sleepy", expertise = "Pillows"}).save()
+    
+    # Create some Pilot Persons
+    D = PilotPerson(name = "Porco Rosso", airplane = "Savoia-Marchetti").save()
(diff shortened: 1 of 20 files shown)


robinedwards:
Hey thanks for this, haven't had time to have a complete read through but there is a number of test failures which need investigating. Have you tried running the test suite via nose? Your setup method isn't required and is already handled.

robinedwards:
Regarding point 1, I think this case warrants its own exception class which is more informative than just a KeyError to the user and definitely needs some docs surrounding it.

Re point 2, I would lean on raising an exception it could highlight other things like the schema the  doesn't match the defined schema in the code. 

My memory is a bit hazy but do we actually save the labels of the parent classes to the node? I think this may need updating somewhere

aanastasiou:
Hello @robinedwards 

Although I did `export` the bolt URL as per the guidelines prior to testing, all 38 tests were failing on my machine because:

`ERROR: Failure: ServiceUnavailable (Cannot acquire connection to Address(host='localhost', port=7687))`

Which I thought was strange. That particular error went away when I added the `neomodel.db.set_connection`. So this is how that line came about.

I did figure that setting up and cleaning up was taken care of by the testing but when my tests seemed to fail, the nodes lingered on the database and this is why I thought I would add these functions to take care of them explicitly. In general, anything apart from the code in the `test_` functions was put there in response to the traces I was getting from `nosetests`.

All the best

robinedwards:
> To keep this functionality consistent, relationship.StructuredRel.start_node and relationship.StructuredRel.end_node had to also be modified so that they instantiate nodes of the correct type as inferred by the labels retrieved from the database (as opposed to simply looking at the type set when defining the model).

So this bit I don't quite agree with in this situation the user has defined two classes and rel type and is expecting one of those two classes. We should perhaps confirm that the node we retrieve has that label but we shouldnt automatically inflate based on label here. But perhaps I am wrong here...

robinedwards:
Regarding tests I am wrong we now use pytest, my apologies! https://github.com/neo4j-contrib/neomodel/commit/99a951f76e9bd40efe69fa7fa5adf3bca50defb7 It would be nice to not duplicate that setup work though and theres built in methods for wiping the db

robinedwards:
I also forgot to say thank you for the PR :+1:

aanastasiou:
> My memory is a bit hazy but do we actually save the labels of the parent classes to the node? I think this may need updating somewhere

This is affected in part by `_abstract_node_` but as per previous discussions, i feel that it is better to leave the complete history of labels with the node as it helps with type inference.

Otherwise, it would have to be ensured that class names are different even across namespaces.

About `start_node` and `end_node`, the thinking was that the end points now are also subject to the same typing constraints. If you specify an end-point with a motherclass, you cannot really be sure what node type will end up on a given `id` (Maintaining all labels is useful at this point). Also, at that point I was not sure if `start_node`, `end_node` are used in `Traversal`s and did not want to risk breaking the "chain".

>Regarding tests I am wrong we now use pytest...

No worries, can adjust as required.

> I also forgot to...

Thank you too, `neomodel` is the better option amongst the alternatives and after figuring out how it sets up metaclasses to do data validation, the temptation to write something from scratch was very big, but I could not see how a re-write could end up vastly different to what it looks like here.

Perhaps in the future it might also be nice to interrogate `neomodel` for the intended changes in order to collect the queries that are to be executed and execute them in batch mode to speed things up a little.

All the best

doc/source/relationships.rst
```diff
@@ -70,6 +70,111 @@ Relationship Uniqueness
...
+        The base actor is defined by its name and a `friends_with` 
+        relationship
+        """
+        name = neomodel.StringProperty(required = True, unique_index = True)
+        friends_with = neomodel.RelationshipTo("BasePerson", "FRIENDS_WITH", model = PersonalRelationship)
+        
+    class TechnicalPerson(BasePerson):
+        """
+        A Technical person specialises BasePerson by adding their 
+        expertise
```
funkyfuture: there are trailing dots missing in this docstring and the following.

doc/source/relationships.rst
```diff
@@ -70,6 +70,111 @@ Relationship Uniqueness
...
+        
+This will show two friends connected with node "Grumpy", one of which is a ``TechnicalPerson`` 
+and the other a ``PilotPerson``.
+
+
+Automatic class resolution
+--------------------------
+
+Neomodel is able to transform nodes to objects, automatically, via a *node-class registry* 
+that is progreessively built up during the definition of the models.
```
funkyfuture: 3rd word has a surplus 'e'.

doc/source/relationships.rst
```diff
@@ -70,6 +70,111 @@ Relationship Uniqueness
...
+        print(some_friend)
+        
+This will show two friends connected with node "Grumpy", one of which is a ``TechnicalPerson`` 
+and the other a ``PilotPerson``.
+
+
+Automatic class resolution
+--------------------------
+
+Neomodel is able to transform nodes to objects, automatically, via a *node-class registry*
```
funkyfuture: commas around 'automatically' add no meaning here.

doc/source/relationships.rst
```diff
@@ -70,6 +70,111 @@ Relationship Uniqueness
...
+
+The dictionary provides a mapping from the set of labels associated with a node to the class 
+that is implied by this set of labels. In the example above, the *node-class registry* would contain 
+the following entries:
+
+    * BasePerson                  --> BasePerson
+    * BasePerson, TechnicalPerson --> TechnicalPerson
+    * BasePerson, PilotPerson     --> PilotPerson
+    
+This automatic resolution is effected by ``neomodel.Database.cypher_objectaware_query`` which is
```
funkyfuture: effected -> affected

doc/source/relationships.rst
```diff
@@ -70,6 +70,111 @@ Relationship Uniqueness
...
+that is implied by this set of labels. In the example above, the *node-class registry* would contain 
+the following entries:
+
+    * BasePerson                  --> BasePerson
+    * BasePerson, TechnicalPerson --> TechnicalPerson
+    * BasePerson, PilotPerson     --> PilotPerson
+    
+This automatic resolution is effected by ``neomodel.Database.cypher_objectaware_query`` which is 
+invoked automatically where needed but it can also be invoked in exactly the same way as ``neomodel.
+Database.cypher_query``, manually, wherever this functionality may be needed.
```
funkyfuture: meaningless commata around 'manually'

doc/source/relationships.rst
```diff
@@ -70,6 +70,111 @@ Relationship Uniqueness
...
+
+Doing so however, requires a bit of caution. As a consequence of the way the *node-class registry* is 
+built up and used, if ``BasePerson``, ``TechnicalPerson``, ``PilotPerson`` were defined 
+in separate files / modules and one of them was not loaded prior to a query to the database, then the 
+*node-class registry* would be unaware of the labels that leads to that particular class and would raise 
+``neomodel.exception.ModelDefinitionMismatch``.
+
+If the exception is not handled, it produces an error message that returns the labels of the node that 
+was retrieved from the database as well as the current *node-class registry*. These two pieces of 
+information can be used to debug the model mismatch further.
```
funkyfuture: if i understand correctly, the use of 'mismatch' seems unappropriate in the designated context. shouldn't it be rather have a notion of 'undefined', 'unknown' or 'not found'?

neomodel/exception.py
```diff
@@ -1,6 +1,28 @@
...
+    nodes for which class definitions do exist but have not been loaded 
+    (or rather imported) or because the retrieved nodes contain more 
+    labels for a known class.
+    
+    In either of these cases the mismatch must be reported
+    """
+    
+    def __init__(self, db_node, current_node_class_registry):
+        """
+        Initialises the exception with the database node that failed to resolve
```
funkyfuture: this sentence doesn't express anything that isn't obvious and i'd consider it boilerplate.

neomodel/exception.py
```diff
@@ -1,6 +1,28 @@
 import importlib
 
 
+class ModelDefinitionMismatch(Exception):
+    """
+    Raised when it is impossible to resolve a Neo4j driver Node to a
```
funkyfuture: i had some time getting this, and would propose this alternative:

> Raised when it is impossible to resolve a node from a Neo4j response to a 
defined model. This can either be caused by models that haven't been imported yet, or because
the a returned node has properties that are not defined in the model.

and as noted regarding `relationships.rst` the exception's name seems to describe only the second of its possible causes.

neomodel/exception.py
```diff
@@ -1,6 +1,28 @@
...
 
+class ModelDefinitionMismatch(Exception):
+    """
+    Raised when it is impossible to resolve a Neo4j driver Node to a 
+    specific object. This can happen as a result of a query returning 
+    nodes for which class definitions do exist but have not been loaded 
+    (or rather imported) or because the retrieved nodes contain more 
+    labels for a known class.
+    
+    In either of these cases the mismatch must be reported
```
funkyfuture: what should this tell me as a user?

neomodel/match.py
```diff
@@ -408,10 +408,8 @@ def _contains(self, node_id):
 
     def _execute(self):
         query = self.build_query()
-        results, _ = db.cypher_query(query, self._query_params)
-        if results:
-            return [self._ast['result_class'].inflate(n[0]) for n in results]
-        return []
+        results, _ = db.cypher_objectaware_query(query, self._query_params)
+        return results
```
funkyfuture: rather `return db.cypher…[0]`?

neomodel/exception.py
```diff
@@ -1,6 +1,28 @@
...
+    
+    def __init__(self, db_node, current_node_class_registry):
+        """
+        Initialises the exception with the database node that failed to resolve 
+        to an object
+        
+        :param db_node: Neo4j driver node object
+        :param current_node_class_registry: Dictionary that maps frozeset of node labels to model classes
+        """
+        Exception.__init__(self, "Node with labels {} does not resolve to any of the known objects\n{}\n".format(",".join(db_node.labels), "\n".join(list(map(lambda x:"{} --> {}".format(",".join(x[0]),x[1]),current_node_class_registry.items())))))
```
funkyfuture: i'd propose this for 'correct' calling of a base class and readability:

```python
super(ModelDefinitionMismatch, self).__init__(
    self, "Node with labels {} does not resolve to any of the known objects\n{}\n".format(
        ",".join(db_node.labels), 
        "\n".join(x[0] + ' --> ' x[1] for x in current_node_class_registry.items())
    )
)
```

neomodel/util.py
```diff
@@ -37,13 +37,28 @@ def clear_neo4j_database(db):
...
+        # Maintains a lookup directory that is used by cypher_objectaware_query
+        # to infer which class to instantiate by examining the labels of the
+        # node in the resultset.
+        # _NODE_CLASS_REGISTRY is populated automatically by the constructor
+        # of the NodeMeta type.
+        self._NODE_CLASS_REGISTRY = {}
 
     def set_connection(self, url):
+        """
+        Sets the connection URL to the address a Neo4j server is set up at
```
funkyfuture: missing trailing dot in several docstrings in this file.

test/conftest.py
```diff
@@ -12,6 +12,7 @@
 
 try:
     clear_neo4j_database(db)
+
```
funkyfuture: this is uncommon and i see no readability gain in this context.

test/test_relationship_models.py
```diff
@@ -6,7 +6,6 @@
 from neomodel import (StructuredNode, StructuredRel, Relationship, RelationshipTo,
                       StringProperty, DateTimeProperty, DeflateError)
 
-
```
funkyfuture: doesn't flake8 complain? well, i do! ;-)

neomodel/exception.py
funkyfuture: on a second thought: there could be two exceptions, one for each case. this should make debugging easier.

doc/source/relationships.rst
aanastasiou: No. "This automatic resolution is effected..." is the intention. The function that provides the ability to do the automatic resolution is `neomodel.Database.cypher_objectaware_query`.

doc/source/relationships.rst
aanastasiou: "...seems unappropriate" --> "...seems inappropriate..."

The model mismatch refers to the condition of the OGM encountering a node that bears more or less labels than what its definition implies.

For example, in the hierarchy "Pilot extends Person", nodes of type "Pilot" are expected to bear the labels "Pilot" and "Person".

If a node was found that contained "Pilot, Person, FighterPilot" then that would be a mismatch with the definitions of the model.

neomodel/exception.py
aanastasiou: This was one of the points in discussion with Robin Edwards about what the behaviour should be and having one exception was the current consensus at the time of writing. The condition is "a different set of labels were encountered while trying to retrieve a node". It can arise in different situations. I think that a single exception can describe the problem well enough.

neomodel/exception.py
aanastasiou: Indeed.

PR #310, opened 2018-03-12
Adds an update method to StructuredNode and other stuff

funkyfuture:
unless i oversaw something there was no functionality of bulk setting of  a node's properties which i found to be useful.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,6 @@
+Version 3.2.6
+ * neomodel.properties.NormalClass is renamed to NormalizedClass
+
 Version 3.2.5
  * Remove use of START in match engine breaking neo4j 3.2
 
@@ -80,7 +83,7 @@ Version 3.0.0 2016-11-27
  * Add __repr__ and __str__ methods to StructuredNode
  * Add get_or_none method to NodeSet
  * Fixed filters persisting across node relationship queries (#208)
- * Un-deprecate is_connected, its intuitive 
+ * Un-deprecate is_connected, its intuitive
  * Fixed NodeSet index returning a list now returns just the node, e.g
  jim.friends[0] returns a node as opposed to a list containing one node.
  * Added missing filter and exclude methods to rel manager enabling: jim.friends.filter(name='bob')
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -36,8 +36,9 @@ import alabaster
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
+    'alabaster',
     'sphinx.ext.autodoc',
-    'alabaster'
+    'sphinx.ext.intersphinx',
 ]
 
 # Add any paths that contain templates here, relative to this directory.
@@ -278,3 +279,9 @@ texinfo_documents = [
 
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
+
+
+# -- Options for intersphinx extension -----------------------------------------
+
+intersphinx_mapping = {'py2': ('https://docs.python.org/2', None),
+                       'py3': ('https://docs.python.org/3', None)}
--- a/doc/source/properties.rst
+++ b/doc/source/properties.rst
@@ -17,32 +17,32 @@ The following properties are available on nodes and relationships:
 Defaults
 ========
 
-*Default values* you may provide a default value to any property, this can also be a function or any callable::
+*Default values* you may provide a default value to any property, this can
+also be a :term:`function` or any callable::
 
         from uuid import uuid4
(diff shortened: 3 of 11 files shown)


funkyfuture:
btw, i nodiced that the docs aren't updated on rtd.

funkyfuture:
can you provide me a link to the `update` in Django that you mean?

actually i thought that properties are validated on setting, but that seems wrong. 
while i think that is desirable, i also think that would be an extra api-refactoring as normalization and validation (that are different, separate things!) would have to be dispatched by `__setattr__` while in- and deflating stays a `save`-related issue. correct? shall i open an issue to discuss this?

funkyfuture:
> Its actually implemented on querysets

but this functionatlity is not intended with this pr. my use-case is to update several properties of a single node from the result of a parsing function.

> I guess ideally an update method should let you update a single property of the node.

the implementation does if it gets a dictionary with one item.

> Regarding read the docs yes it does need updating, I try and do it with every release to PyPi but really looking for someone else to take that over

usually the 'latest' branch on rtd matches the 'master'  branch of a git repo. hence the 'latest' docs should be built on every update un the master branch. i encountered problems with this once, but don't remember details atm. are there builds in the build history that reflect pushes/merges to the master branch recently?

robinedwards:
Yea I see your point however I think there is a case for updating just one property.

I definitely think at the very least update() should persist its changes. I dont really see a need for update_and_save().

See also my note about Normalized

funkyfuture:
> Yea I see your point however I think there is a case for updating just one property.

i still don't get it. there are two ways to update a single property:

    node.some_property = property_value

or:

    node.update({'some_property': property_value})

what am i missing?

> I definitely think at the very least update() should persist its changes. I dont really see a need for update_and_save().

but then there'd be unneeded database transactions when this method is used more than one time before the logic considers its changes 'complete'. the `update_and_save` method was intended purely for convenience as i'd anticipate it wouldn't be rare that folks add such via subclassing.

however, i have to admit i'm not very happy with reserving more possible property names with method names ((`category`,) `update` and `labels` seem likely to be wanna used as proprty name). maybe all these methods should also be available via a manager object that is bound to an instance (like `nodes`, but only for instance related attributes)?

> See also my note about Normalized

i don't [see any](https://github.com/neo4j-contrib/neomodel/pull/310/commits/4b84177a5ceedf7ce672dfe4638c17534b9d77da). was that concerning the test whose changes i missed to commit first?

robinedwards:
Well for that I agree, 'category' has been deprecated for a long time now so perhaps it can be 'unreserved' as a property name

Sorry maybe I am not being clear. I see the update method as a way of optionally updating one or more properties on a node in a single operation.

```python
node.some_property = value
node.save() # updates all of the nodes properties in the database

# updates only the passed properties on the node in the database
node.update({some_property: 1, some_other_property: 2})
```

Regarding Normalized rename I am referring not to within in neomodel code but other people using the library. The class its self was contributed by a member of the community and I know they are using it so we cant rename without providing a deprecation warning of the old name.

funkyfuture:
i come to the impression that we have differerent understandings of `update`, where you're rather using the meaning from a database context while i do from Python's `dict.update`. i'm using it the latter way as i throw the results of several parsing functions on it - and later do a single database update / `save`.

if my suspicion is the case, this design leads to confusion, i can think of these alternatives:

1. use another method name (`update_properties`?).
2. having a proxy object bound to a node that implements all mapping methods to query and manipulate. (e.g. `node.properties.update`)
3. having a manager object to 'unreserve' most of a node's namespace. 1. would need to be addressed anyway in this case.

i'll amend the deprecation warning later this day.

robinedwards:
Right I see where you are coming from. I had never thought about it that way and it is in fact more Pythonic.

I wonder about making update() return self to allow ```update({foo: bar}).save()```? We also have __properties__ for accessing the nodes properties as a dict, so we could support update here but perhaps thats a bit long winded?

funkyfuture:
> We also have properties for accessing the nodes properties as a dict, so we could support update here

thanks for the hint, that'd be absolutely the most consistent approach imo.

i like the idea regarding calling `save` on a result. it seems 'unpure' though. what about this?: the `properties` object gets a `save` method that delegates to the one of the node, and `properties.update` returns the properties object. so: `node.properties.update({foo: bar}).save()`

robinedwards:
Yea you mean re-use the existing one ```node.__properties__```?

funkyfuture:
a'ight, i amended the following:

1) added deprecation info regarding `NormalProperty` (and finally understood my cryptic node ;-)
2) added a commit with improvements to use `match.Traversal`
3) removed the commit that added the `update` method to move that to another pr.

the last is motivated as i guess this involves bigger changes. hence this pr should be ready to merge.

PR #311, opened 2018-03-16
Respect db_property on model properties in get_or_create

mprahl:
When calling `get_or_create` on a node that already exists and has a property with `db_property` set, I would get the error `Cannot merge node using null property value for var_name`, which is now addressed in `core.py` by checking to see if `db_property` is set on the property.


mprahl:
I added a test that reproduces the issue.

mprahl:
@robinedwards may you please review this PR? The bug this PR resolves is a blocker for my use-case.

robinedwards:
Thats great thank you!

PR #312, opened 2018-03-16
Respect when db_property is set in install_labels

mprahl:
Currently, `neomodel_install_labels` uses the wrong property name when a `UniqueIdProperty` has `db_property` set. This PR fixes that.

Reproducer:
```python
class ContainerBuilds(StructuredNode):
    id_ = UniqueIdProperty(db_property='id')
```
Then run `neomodel_install_labels` on the file. It will create the unique constraint on `id_` instead of `id`.

### DIFF
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -84,22 +84,23 @@ def install_labels(cls, quiet=True, stdout=None):
         return
 
     for key, prop in cls.defined_properties(aliases=False, rels=False).items():
+        db_property = prop.db_property or key
         if prop.index:
             if not quiet:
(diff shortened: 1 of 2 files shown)


funkyfuture:
adding your 'Reproducer' as a test would be a good thing.

mprahl:
@funkyfuture sure thing. I added a test to reproduce the issue.

mprahl:
@robinedwards may you please review this PR? The bug this PR resolves is a blocker for my use-case.

robinedwards:
Thank you again!

Issue #287, opened 2017-11-27
@transaction decorator ultimately tries to resolve a hostname


gegenschall:
The `@neomodel.db.transaction` decorator tries to connect to a Neo4j database. If you cannot make a Neo4j database available while importing code that uses the decorator, you get connection errors.

I got bitten by this when trying to build a Docker image that uses Django management commands, which in turn try to import code that uses the transaction decorator. I am basically unable to build Docker images when using the decorator because the host running the Neo4j database is neither known (i.e. can be resolved) nor available. I have to resort to using some clumsy try-except hell instead.

robertlagrant:
Yes - I have the same problem:

```
  File "/Users/user/Documents/GitHub/project/blueprint/controller.py", line 16, in <module>
    @db.transaction
  File "/Users/user/Documents/GitHub/project/venv/lib/python3.6/site-packages/neomodel/util.py", line 27, in wrapper
    self.set_connection(config.DATABASE_URL)
  File "/Users/user/Documents/GitHub/project/venv/lib/python3.6/site-packages/neomodel/util.py", line 61, in set_connection
    max_pool_size=config.MAX_POOL_SIZE)
  File "/Users/user/Documents/GitHub/project/venv/lib/python3.6/site-packages/neo4j/v1/api.py", line 112, in driver
    return driver_class(uri, **config)
  File "/Users/user/Documents/GitHub/project/venv/lib/python3.6/site-packages/neo4j/v1/direct.py", line 63, in __init__
    pool.acquire()
  File "/Users/user/Documents/GitHub/project/venv/lib/python3.6/site-packages/neo4j/v1/direct.py", line 44, in acquire
    raise ServiceUnavailable("Cannot acquire connection to {!r}".format(self.address))
neo4j.exceptions.ServiceUnavailable: Cannot acquire connection to Address(host='localhost', port=7687)
```

`db` could defer connecting until a query is run?

PR #313, opened 2018-03-20
Ensure_connection only called when accessing the database, not when c…

robertlagrant:
…onstructing the transaction decorator - see #287 

This change is to allow neomodel-enabled apps to start when a database is not present, and it will only throw a connection error when attempting to access the database. This removes the need for separate error handling when the database starts after the app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ development.env
 .python-version
 .ropeproject
 \#*\#
+.eggs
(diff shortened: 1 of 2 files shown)


manoadamro:
Amazing!

robinedwards:
Fantastic thank you

robertlagrant:
@robinedwards is there a new version scheduled with the latest PRs (including this one, natch :))? Be good to know if we'll be able to just increment a version number or we have to look at installing directly from Git.

Thanks :-)

PR #315, opened 2018-03-20
A proposal for a Python 3-only / a next major release

funkyfuture:
as intended with #310 i desired an interface to the properties of a node or a relationship that behaves like a mapping. @robinedwards pointed on the `__properties__` property, but this one is immutable.

on my way to a solution i came around a lot of code that looked over-complicated (as in grown with the years) and a usage of meta-classes that didn't mangle out shit of the model classes as it could have. as a result the introspective properties are extended and a lot of stuff is only calculated once now.

addressing these soon lead to the point where maintaining code for both Python 2 and 3 became impossible, so i opted for the full elegance of new-style classes.

another major point was that it was hard to understand code that is scattered across many modules (and the contained pieces relate close to each other) with runtime imports to address cross dependencies. i cracked that nut by adding a typing system and regrouping of related module contents.

in the end i am surprised how much performance improvement this *seems* (no laboratory measurements here) to bring. with a current use-case i'm working on, a data import procedure reduced from ~2:30min to ~2:15min, that'd be ~10%. neomodel's test suite runtime drops from ~80s to ~70s on my local machine.

no public api changes are introduced so far, but a breaking release has this opportunity. e.g. i think of a class kw argument `abstract` instead of the class property `__abstract_node__`.

a lot of effort can nonetheless still be put into code readability and style, a look on the test suite's coverage and especially the documentation (it isn't as good as the library deserves, imo.).


robinedwards:
This is fantastic work Frank, I'd love to include this in the 4.x series please see my email

funkyfuture:
i tried to rebase this branch on master and of course it didn't turn out well. to achieve that it'd be probably easier to cherry-pick all commits from this branch onto a new one.

aanastasiou:
@funkyfuture Would it be possible to indicate which commits from this PR should unreservedly feature in an upcoming major release to start preparing for it?

aanastasiou:
@funkyfuture Any updates on this?

PR #317, opened 2018-03-26
Add missing parameter in Relationships documentation

mprahl:
Using the documentation as is provides the error `TypeError: RelationshipFrom() takes at least 2 arguments (2 given)`.

### DIFF
--- a/doc/source/relationships.rst
+++ b/doc/source/relationships.rst
@@ -20,10 +20,10 @@ It's possible to (softly) enforce cardinality restrictions on your relationships
 Remember this needs to be declared on both sides of the definition::
 
     class Person(StructuredNode):
-        car = RelationshipTo('Car', 'CAR', cardinality=One)
+        car = RelationshipTo('Car', 'OWNS', cardinality=One)
 
     class Car(StructuredNode):
-        owner = RelationshipFrom('Person', 'CAR', cardinality=One)
+        owner = RelationshipFrom('Person', 'OWNS', cardinality=One)
 
 The following cardinality classes are available:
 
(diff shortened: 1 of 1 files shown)


doc/source/relationships.rst
```diff
@@ -23,7 +23,7 @@ Remember this needs to be declared on both sides of the definition::
         car = RelationshipTo('Car', 'CAR', cardinality=One)
 
     class Car(StructuredNode):
-        owner = RelationshipFrom('Person', cardinality=One)
+        owner = RelationshipFrom('Person', 'CAR', cardinality=One)
```
funkyfuture: wouldn't this rather be `'OWNED_BY` or similar?

funkyfuture: afaik, that argument sets the [relationship type](https://neo4j.com/docs/developer-manual/current/terminology/#term-relationship-type). the relationship source is set when the model class is initialized.

mprahl: @funkyfuture yes you're right. Let me fix that.

mprahl:
@funkyfuture so I tried your suggestion and it created a relationship from `Person` to `Car` called `OWNED_BY`. Here is a screenshot of a call to `db.schema()`:
screenshot from 2018-03-27 09-04-36

I had it right the first time, so I'll revert back to what I had originally.

doc/source/relationships.rst
mprahl: As mentioned, in the comments, I had this right originally. The relationship name should be `CAR` because we want to access the `CAR` relationship from the `Person` node.

funkyfuture:
why don't you change the relationship type from a passive to an active? A `Person` `OWNS` a `CAR`.

'Person car car' is hardly a meaningful expression.

mprahl:
@funkyfuture I agree but this is what documentation had to begin with. I'll make this change in a separate commit so the maintainer can decide what they prefer.

funkyfuture:
i'm not sure whether this is docuemnted, but the relationship type hardly plays a role in the model sphere. to get the owner of a car: `car.owner.single()`

PR #321, opened 2018-03-30
Remove a weird import with a TODO

henryfjordan:
The import in question is done at the top of the file, so this import is unecessary


henryfjordan:
Not sure why the tests failed, some intermittent build error on:
`sudo apt-get update && sudo apt-get install oracle-java8-installer`

PR #322, opened 2018-04-05
Update version in __init__.py

manoadamro:


neomodel/__init__.py
```diff
@@ -18,4 +18,4 @@
 __email__ = 'robin.ge@gmail.com'
 __license__ = 'MIT'
 __package__ = 'neomodel'
-__version__ = '3.2.4'
+__version__ = '3.2.6'
```
funkyfuture: it'd be actually preferable to `= pkg_resources.get_distribution('neomodel').version`.

manoadamro: That's a much better idea!

robinedwards:
Wow I didnt know about this. Good call @funkyfuture

PR #325, opened 2018-04-14
Add the "first" and "first_or_none" methods on the NodeSet class

mprahl:
I'm used to SQLAlchemy and they have a handy [first](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.first) method. This PR implements something very similar to that but instead matches the behavior of `get` and `get_or_none` for consistency.


robinedwards:
Brilliant thanks @mprahl

PR #326, opened 2018-04-16
Don't raise a cardinality exception when connecting an already connected node

mprahl:
I'd like to start using cardinality in my project but my existing code that imports data into Neo4j attempts to connect nodes without checking if the relationship already exists, which works just fine when cardinality is not used. I'd like to continue using this approach, so this PR will check to see if the node being connected is already connected before raising an `AttemptedCardinalityViolation` exception.

### DIFF
--- a/neomodel/cardinality.py
+++ b/neomodel/cardinality.py
@@ -36,7 +36,7 @@ class ZeroOrOne(RelationshipManager):
         :type: dict
         :return: True / rel instance
         """
-        if len(self):
+        if len(self) and node not in self:
             raise AttemptedCardinalityViolation(
                     "Node already has {0} can't connect more".format(self))
         else:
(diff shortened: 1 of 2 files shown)


robinedwards:
So this is a tricky one, it depends on your data model, I believe neo4j allows multiple connections between the same node so that's the reason this exception is thrown even if that node is already connected..

mprahl:
@robinedwards Ah I see. Any suggestions other than doing this manually in my code before creating a relationship?

robinedwards:
I guess it would require a major revision but we could have two different cardinality classes to represent these differing behaviours. You could also subclass the relationship manager in your own codebase to augment this behaviour. Not sure if the code is particularly well structured for doing this at the moment.

mprahl:
@robinedwards I'll work around this in my code. Thanks for the review and help!

PR #327, opened 2018-04-17
Add the "disconnect_all" and "replace" methods on relationship properties

mprahl:
My project that uses neomodel requires replacing all existing relationships with a new one. I found myself doing this by looping through all the connected nodes and disconnect the relationships individually. This was repetitive across the code base and inefficient. This PR adds the ability to replace all relationships of a specific type with a new one. It also adds the ability to remove all existing relationships of a specific type.


robinedwards:
Brilliant thank you!

PR #328, opened 2018-04-17
Lower the required pytz version

mprahl:
neomodel is packaged in EPEL (Extra Packages for Enterprise Linux) for EL7 (RHEL 7/CentOS 7) and the latest version of pytz available on EL7 is 2016.10. Having the version of pytz too high makes it so that neomodel can't be packaged for EL7 and current versions of Fedora (Fedora uses version 2017.2).


robinedwards:
Ok fair, thanks.

PR #336, opened 2018-05-15
Add support for query filter with OR # 332

mastisa:
Add support to query with OR logical operator for example if or_created__gte=30, updated_lte=30, and_name='saeed' the where query will be like created >= 30 OR updated <= 30 AND name='saeed'. I use or_ , and_ for OR and AND. If there is no or_ or and_, I will use and for backward compatibility.  Every logical operator will use to perform operand on current property and next property. If there is no next property, logical operator will be unused.

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -106,6 +106,12 @@ OPERATOR_TABLE = {
 # add all regex operators
 OPERATOR_TABLE.update(_REGEX_OPERATOR_TABLE)
 
+# list all supported operators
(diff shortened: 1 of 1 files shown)


robinedwards:
Hey thanks for the contribution however I think id be more inclined to follow djangos Q model for achieving this as I think the approach you have suggested has limited flexibility.

juanmhidalgo:
@robinedwards I added a PR (#360 ) using the same aproach as Django's Q model, can you take a look?

Issue #335, opened 2018-05-15
neo4j.exceptions.ClientError: No write operations are allowed directly on this database. Writes must pass through the leader. The role of this server is: FOLLOWER


robertlagrant:
A funny one:

- We have a 3-node causal-clustered neo4j setup
- I've changed the routing protocol to be `bolt+routing`
- We're using Neomodel with `@db.transaction`

We're getting intermittent errors as per the issue title - i.e. it's trying to write to a follower node, and presumably bolt+routing isn't sending the transaction to the leader.
Am I missing something? Is it that if the first interaction with the database is a read, that it opens the transaction on a follower node? Can I force it to the leader for every transaction?

aanastasiou:
@robertlagrant Would it be possible to share a little bit more information on your cluster configuration? Is that supposed to be 3 CORE servers? There are some conditions where what you describe might be the intended behaviour at least as far as RAFT is concerned (i.e. see [this](https://neo4j.com/docs/operations-manual/current/clustering/lifecycle/)). I am trying to see how much of this can be dealt with at the level of `neomodel` and how much of this is external to it.

mvanderkroon:
@aanastasiou I believe so. I have forked the repo, made the necessary changes and would be quite happy to issue a pull request. Should I point it to your `master` branch?

aanastasiou:
@mvanderkroon Thank you very much and I do not see why not. It should be sent as a pull request to the main `neomodel` repo. All the best.

aanastasiou:
@robertlagrant Thank you for your response, I think that the discussion with @mvanderkroon on the pull request was very informative about the specifics.

kant111:
Why follower cannot accept writes?

robertlagrant:
@kant111 because that's not how Neo4J works.

ayoubelmimouni:
when using a connection URL of bolt+routing:// this indicates the session is now cluster aware, whereas bolt:// does not understand the other members in a cluster.
However it is not simply the bolt+routing:// connection URL is only half the story. It is also the usage of session.readTransaction() and session.writeTransaction() whereby each allows you to pass the Cypher to be executed. If you send a cypher statement through session.writeTransaction and the connection URL was bolt+routing:// then regardless of the member connected to, the Cypher write statement will be routed to the LEADER. As such if one connects to bolt+routing://<followerIP> and calls a session.writeTransaction() as the transaction is defined as a write it will automatically be routed to the LEADER.
[...]

PR #337, opened 2018-05-16
Explicit write transaction mode

robertlagrant:
Causal clustering has a leader and multiple followers. Only leader can be written to.

Initiating a transaction with `@db.transaction` doesn't seem to allow the specifying of a transactionality type (Bolt is READ or WRITE), thus it's possible that the `bolt+routing` protocol could route a write transaction to a follower.

This PR allows the developer to choose `@db.write_transaction`, which will set `access_mode` to `"WRITE"`, directing Bolt to the leader of a causal cluster. It defaults to the previous behaviour to avoids backwards compatibility issues.

### DIFF
--- a/doc/source/transactions.rst
+++ b/doc/source/transactions.rst
@@ -1,6 +1,6 @@
 Transactions
 ------------
-transactions can be used via a context manager::
+Transactions can be used via a context manager::
 
     from neomodel import db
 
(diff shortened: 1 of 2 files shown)


robertlagrant:
This is to address issue #335

robertlagrant:
@robinedwards no problem - I've added a bit of documentation; let me know what you think.

robinedwards:
This great, really appreciate it @robertlagrant

robertlagrant:
@robinedwards it's a pleasure! Is there a release scheduled for some time soon? I'm trying to figure out if I need a custom build or not.

robinedwards:
Released!

PR #339, opened 2018-05-17
add max length check to string property

Lazy-Y:
add max length check to string property


robinedwards:
Thanks for this.

I think the max_length property needs to live on the actual string property as it may not necessarily apply to other properties inheriting from NormalProperty?

Issue #340, opened 2018-05-24
There is no way to connect to multiple databases from a single process.


vaibhavmoengage:
USE CASE: We have two neo4j clusters. We want to redirect data to different cluster on the basis of certain keys. The decision to redirect data is to be taken from within the process. So i need a way to create a database connection which takes a input parameter as url.

reorx:
may I ask why this issue was closed? I still can't find any way to connect to multiple databases in a single process

PR #341, opened 2018-05-24
way to plugin url in database object added

vaibhavmoengage:


vaibhavmoengage:
Issue: https://github.com/neo4j-contrib/neomodel/issues/340

PR #347, opened 2018-06-14
Support neo4j-driver v1.6.0

mprahl:
In the v1.6 of the Python Neo4j driver, `properties` got renamed to `_properties`. This PR adds a compatibility for both ways.


robinedwards:
Thanks @mprahl much appreciated

PR #355, opened 2018-07-11
Raise ValueError when refreshing a non existent inflated node

### LINKED ISSUE #354: .refresh() on an inflated node that does not exist raises IndexError
This comes from following line (core.py/495), where the structure of the cypher response is unchecked:
```
            node = self.inflate(self.cypher("MATCH (n) WHERE id(n)={self}"
                                            " RETURN n")[0][0][0])
```

A more appropriate exception should be raised, for instance ValueError or DoesNotExist.

lerela:
Fixes #354

### DIFF
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -492,8 +492,11 @@ class StructuredNode(NodeBase):
         """
         self._pre_action_check('refresh')
         if hasattr(self, 'id'):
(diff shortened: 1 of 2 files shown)


robinedwards:
I think theres a DoesNotExist exception we could raise here which is more specific

lerela:
@robinedwards I updated the PR.

robinedwards:
Super thanks!

Issue #332, opened 2018-04-29
Filter with OR


sbabashahi:
Recently I start to useing neo4j and neomodel and now I need to filter 2 parameters with OR in neomodel. I searched in document and codes but couldn't find anything about it. In django model there is a Q object to handle OR in queries. How can handle OR query in neomodel?

cr17ms:
I made some changes to solve this issue in #336.

MardanovTimur:
close this issue

PR #360, opened 2018-08-11
Added support for Q() in filter and exclude .

juanmhidalgo:
Implemented Q() filter and exclude like Django.
(Filter with OR #332)

### DIFF
--- a/doc/source/queries.rst
+++ b/doc/source/queries.rst
@@ -54,6 +54,70 @@ The filter method borrows the same django filter format with double underscore p
 - regex - matches a regex expression
 - iregex - matches a regex expression, case insensitive
 
+Complex lookups with ``Q`` objects
+==================================
+
+Keyword argument queries -- in `filter`,
+etc. -- are "AND"ed together. If you need to execute more complex queries (for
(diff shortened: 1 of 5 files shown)


juanmhidalgo:
I modified tree.py to force CI to run the tests again.
It would be great if you can review this.

robinedwards:
Hey thank you for this its a really useful feature. Can I request that the `Node` class that's borrowed from Django is renamed as people could potentially find this confusing. Perhaps just merge the methods into Q or name it something like QBase?

robinedwards:
Also rename the tree package to `match_q.py` or something similar

juanmhidalgo:
@robinedwards Done.

robinedwards:
Thank you thats brilliant!

robinedwards:
Sorry one last request would you be able to add a few example queries in https://github.com/neo4j-contrib/neomodel/blob/master/doc/source/queries.rst so other people are aware of the feature?

juanmhidalgo:
@robinedwards 
The documentation is ready. I took the Django documentation for Q objects and modified a little bit.

robinedwards:
Thnk you @juanmhidalgo !

PR #362, opened 2018-08-31
Empty filter bug

MardanovTimur:
If you're trying do this
```
SomeModel.nodes.filter().filter(arg1=val1).all()
```
neomodel throws exception like this
```
MATCH (somemodel:SomeModel) WHERE AND somemodel.arg1={arg1_1} RETURN ....
```
I fixed it. Sorry for my bad english


robinedwards:
Thank you!

PR #364, opened 2018-09-05
Filter by relationship params and get it

MardanovTimur:
```
some_isntance.city.filter_relationsips(some_city_instance, {relationship_filter_kwargs}) -> returns relationships
```

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -125,6 +125,9 @@ def install_traversals(cls, node_set):
         traversal = Traversal(source=node_set, name=key, definition=rel.definition)
         setattr(node_set, key, traversal)
 
+re._alphanum_str = frozenset(
+    "_абвгдеёжзийклмнопрстуфхцчшщъыьэюяАБВГДЕЁЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ\
+            abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890")
 
 def process_filter_args(cls, kwargs):
     """
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -1,4 +1,4 @@
-#encoding=utf-8
+# encoding=utf-8
 from .core import StructuredNode, db
 from .properties import AliasProperty
 from .exceptions import MultipleNodesReturned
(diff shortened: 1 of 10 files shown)


MardanovTimur:
@robinedwards  it'is works too.

MardanovTimur:
image

robinedwards:
Hi thanks for this! There's actually already a method for this called 'match' which I dont particularly like the name. As we are getting close to a 4.0.0 release, We could rename it to filter_relationships and add a deprecation note to the old one?

MardanovTimur:
No, match returns the nodes whose filter by relationships. filter_relationships returns relationships, filtering by rels too

neomodel/match.py
```diff
@@ -125,6 +126,9 @@ def install_traversals(cls, node_set):
         traversal = Traversal(source=node_set, name=key, definition=rel.definition)
         setattr(node_set, key, traversal)
 
+re._alphanum_str = frozenset(
+    "_абвгдеёжзийклмнопрстуфхцчшщъыьэюяАБВГДЕЁЖЗИЙКЛМНОПРСТУФХЦЧШЩЪЫЬЭЮЯ\
+            abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890")
```
robinedwards: I think theres an actual regex property for this

neomodel/match.py
```diff
@@ -264,22 +267,27 @@ def build_traversal(self, traversal):
...
-
-        # build source
+        #build source
         lhs_ident = self.build_source(traversal.source)
-        rhs_ident = traversal.name + rhs_label
-        self._ast['return'] = traversal.name
+        rhs_ident = "{alias}:{def_label}".format(alias=traversal.name,
+                def_label=traversal.definition.get('node_class').__label__)
+
+
```
robinedwards: extra white spaxe

neomodel/match.py
```diff
@@ -264,22 +267,27 @@ def build_traversal(self, traversal):
         """
         traverse a relationship from a node to a set of nodes
         """
-        # build source
-        rhs_label = ':' + traversal.target_class.__label__
-
-        # build source
+        #build source
```
robinedwards: space after #the

neomodel/match.py
MardanovTimur: It does not work with the Cyrillic alphabet. Now it works. Before Cyrillic symbols re.escape insert 2 backslashes and neomodel build incorrect query.

neomodel/match.py
MardanovTimur: Ok, i will fix it now

neomodel/match.py
MardanovTimur: ok

MardanovTimur:
@robinedwards ready

neomodel/match.py
robinedwards: https://docs.python.org/2/library/re.html#re.UNICODE we just need to set this flag where the regex is being used. Can you do that and put it in a separate MR from this one please. Thanks!

robinedwards:
I think `match(some_property='bar').all_relationships()` maybe what you are after?

whatSocks:
will check on in https://github.com/neo4j-contrib/neomodel/issues/594

Issue #380, opened 2018-11-17
Deprecation warning: time.clock()


mjmare:
I'm getting these deprecation warnings running neomodel 3.3.0 when running some tests:

```
=============================== warnings summary ===============================
test_model.py::test_introspection
  /Users/mjm/Dropbox/Devel/_Experiments/Ariadne/venv/lib/python3.7/site-packages/neomodel/util.py:195: DeprecationWarning: time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8: use time.perf_counter or time.process_time instead
    start = time.clock()
  /Users/mjm/Dropbox/Devel/_Experiments/Ariadne/venv/lib/python3.7/site-packages/neomodel/util.py:198: DeprecationWarning: time.clock has been deprecated in Python 3.3 and will be removed from Python 3.8: use time.perf_counter or time.process_time instead
    end = time.clock()
```

aanastasiou:
Indeed, thanks.

By the way, do you think you could have a go at fixing it?

aanastasiou:
Never mind, thank you very much for pointing this out. Will be queued for the next release. All the best

PR #385, opened 2018-11-20
Minor fix to the deprecation warning reported in #380

aanastasiou:
Fixes the deprecation warning raised by the use of `time.clock`, now using `time.process_time()`.


leonaidus:
This fix will not work with python version earlier than 3.3.
To understand try to run:

    time.process_time()

on python v2.7.

aanastasiou:
That's great, thank you.

Issue #303, opened 2018-02-26
Gracefully handled exceptions within a transaction cause the transaction to fail


aanastasiou:
# The problem
This is a sneaky bug because the traceback is all due to the Neo4j Python driver when the issue is more likely due to the way neomodel `TransactionProxy` handles the presence of additional exceptions.

# Reproducing it

```
# Set up a very simple model for the tests
class PersonalRelationship(neomodel.StructuredRel):
    """
    A very simple relationship between two basePersons that simply records 
    the date at which an acquaintance was established.
    This relationship should be carried over to anything that inherits from 
    basePerson without any further effort.
    """
    on_date = neomodel.DateProperty(default_now = True)
    
class BasePerson(neomodel.StructuredNode):
    """
    Base class for defining some basic sort of an actor.
    """
    name = neomodel.StringProperty(required = True, unique_index = True)
    friends_with = neomodel.RelationshipTo("BasePerson", "FRIENDS_WITH", model = PersonalRelationship)
    
class TechnicalPerson(BasePerson):
    """
    A Technical person specialises BasePerson by adding their expertise
    """
    expertise = neomodel.StringProperty(required = True)
```

This simple case works:

```
with neomodel.db.transaction:
        A = TechnicalPerson(name = "Grumpy", expertise = "Grumpiness").save()
```

This however fails:

```
A = TechnicalPerson(name = "Grumpy", expertise = "Grumpiness").save()

with neomodel.db.transaction:
    try:
        A = TechnicalPerson(name = "Grumpy", expertise = "Grumpiness").save()
    except neomodel.UniqueProperty:
        A = TechnicalPerson.nodes.get(name = "Grumpy", expertise = "Grumpiness")
```

The latter snippet fails with a trace that concludes with:
```
. . .
 raise CypherError.hydrate(**metadata)
neo4j.exceptions.ClientError: Transaction rolled back even if marked as successful
```

# Reason & Potential Resolution

`util.Database` defines `transaction` as a property. Upon calling `with neomodel.db.transaction:`, this transaction property getter returns a `util.TransactionProxy` object whose `__enter__` kickstarts the transaction.

If an exception occurs within the `with` block, the `util.Database.TransactionProxy.__exit__` is called with the exception type, value and traceback. 

Neomodel catches the generic `neo4j.v1.CypherError` and specifically the `ConstraintValidationFailed` to raise its own `exception.UniqueProperty`.

[...]

PR #398, opened 2019-01-09
[WIP] Feature/neo retry logic

mvanderkroon:
Adds retry logic to the query execution code which should mitigate high level, transient neo4j errors such as the ones mentioned in #335

### DIFF
--- a/neomodel/exceptions.py
+++ b/neomodel/exceptions.py
@@ -13,6 +13,7 @@ class AttemptedCardinalityViolation(NeomodelException):
 
     Example: a relationship of type `One` trying to connect a second node.
     """
+
     pass
 
 
@@ -22,13 +23,15 @@ class CardinalityViolation(NeomodelException):
 
     For example a relationship type `OneOrMore` returns no nodes.
     """
+
     def __init__(self, rel_manager, actual):
         self.rel_manager = str(rel_manager)
         self.actual = str(actual)
 
     def __str__(self):
-        return "CardinalityViolation: Expected: {0}, got: {1}."\
-            .format(self.rel_manager, self.actual)
+        return "CardinalityViolation: Expected: {0}, got: {1}.".format(
+            self.rel_manager, self.actual
+        )
 
 
 class ModelDefinitionMismatch(NeomodelException):
@@ -41,7 +44,7 @@ class ModelDefinitionMismatch(NeomodelException):
     
     In either of these cases the mismatch must be reported
     """
-    
+
     def __init__(self, db_node, current_node_class_registry):
         """
         Initialises the exception with the database node that failed to resolve 
@@ -53,15 +56,21 @@ class ModelDefinitionMismatch(NeomodelException):
         """
         self.db_node = db_node
         self.current_node_class_registry = current_node_class_registry
-        
+
     def __str__(self):
         node_labels = ",".join(self.db_node.labels)
-        ncr_items = list(map(lambda x:"{} --> {}".format(",".join(x[0]), x[1]),
-                             self.current_node_class_registry.items()))
+        ncr_items = list(
+            map(
+                lambda x: "{} --> {}".format(",".join(x[0]), x[1]),
+                self.current_node_class_registry.items(),
+            )
+        )
         ncr_items_multiline = "\n".join(ncr_items)
 
-        return "Node with labels {} does not resolve to any of the known " \
-               "objects\n{}\n".format(node_labels, ncr_items_multiline)
+        return (
+            "Node with labels {} does not resolve to any of the known "
+            "objects\n{}\n".format(node_labels, ncr_items_multiline)
+        )
 
 
 class ConstraintValidationFailed(ValueError, NeomodelException):
@@ -77,9 +86,12 @@ class DeflateError(ValueError, NeomodelException):
         self.obj = repr(obj)
 
     def __str__(self):
-        return ("Attempting to deflate property '{0}' on {1} of class '{2}': "
-                "{3}".format(self.property_name, self.obj,
-                             self.node_class.__name__, self.msg))
+        return (
+            "Attempting to deflate property '{0}' on {1} of class '{2}': "
+            "{3}".format(
+                self.property_name, self.obj, self.node_class.__name__, self.msg
+            )
+        )
 
 
 class DoesNotExist(NeomodelException):
@@ -113,7 +125,8 @@ class InflateConflict(NeomodelException):
     def __str__(self):
         return """Found conflict with node {0}, has property '{1}' with value '{2}'
             although class {3} already has a property '{1}'""".format(
-            self.nid, self.property_name, self.value, self.cls_name)
+            self.nid, self.property_name, self.value, self.cls_name
+        )
 
 
 class InflateError(ValueError, NeomodelException):
@@ -124,9 +137,12 @@ class InflateError(ValueError, NeomodelException):
         self.obj = repr(obj)
 
     def __str__(self):
-        return ("Attempting to inflate property '{0}' on {1} of class '{2}': "
-                "{3}".format(self.property_name, self.obj,
-                             self.node_class.__name__, self.msg))
+        return (
+            "Attempting to inflate property '{0}' on {1} of class '{2}': "
+            "{3}".format(
+                self.property_name, self.obj, self.node_class.__name__, self.msg
+            )
+        )
 
 
 class DeflateConflict(InflateConflict):
@@ -134,12 +150,13 @@ class DeflateConflict(InflateConflict):
         self.cls_name = cls.__name__
         self.property_name = key
         self.value = value
-        self.nid = nid if nid else '(unsaved)'
+        self.nid = nid if nid else "(unsaved)"
 
     def __str__(self):
         return """Found trying to set property '{1}' with value '{2}' on node {0}
             although class {3} already has a property '{1}'""".format(
-            self.nid, self.property_name, self.value, self.cls_name)
+            self.nid, self.property_name, self.value, self.cls_name
+        )
 
 
 class MultipleNodesReturned(ValueError, NeomodelException):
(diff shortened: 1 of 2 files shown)


aanastasiou:
@mvanderkroon This is great, thank you.

neomodel/util.py
```diff
@@ -193,17 +219,58 @@ def cypher_query(self, query, params=None, handle_unique=True, retry_on_session_
         try:
             # Retrieve the data
             start = time.clock()
-            response = session.run(query, params)
-            results, meta = [list(r.values()) for r in response], response.keys()
+
+            delay = 1
+            while True:
+                last_exception = None
+                cts = time.clock()
```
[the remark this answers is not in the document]
mvanderkroon: In my original snippet (and the hotfix I use locally) I do use `time.time()`, however the original code uses `time.clock()` and I figured there was a reason for it. I'll change it to use `time.time()` as I believe backwards compatibility is desirable

neomodel/util.py
```diff
@@ -193,17 +219,58 @@ def cypher_query(self, query, params=None, handle_unique=True, retry_on_session_
...
+                    raise last_exception
+
+                try:
+                    response = session.run(query, params)
+                    results, meta = (
+                        [list(r.values()) for r in response],
+                        response.keys(),
+                    )
+                    break
+                except Exception as e:
```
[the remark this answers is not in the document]
robertlagrant: Yes, that's right!

I think the neobolt NotALeader exception is a good pick for when the we're now talking to a follower. As a bonus, given how it's named I assume it also will throw an exception when the node is in CANDIDATE status, before the cluster forms.

A good example of (2) is that the app has been directed (say) to node 1, which is the leader, and node 1 dies for some reason. Nodes 2 and 3 re-establish themselves as follower and leader, but while node 1 is physically coming back up it may be at a point where it's not even serving network requests, let alone have cluster info. So if the app tries to connect in that time, it almost certainly will get a connection timeout. When it tries to connect to the old node before it's got the cluster info, that's probably the service unavailable exception. I imagine that neobolt's ConnectionExpired and ServiceUnavailable exceptions will cover those cases.

In the case of ConnectionExpired, it could probably stop trying to connect to that server and connect afresh to the globally configured connection string instead, as that'll probably be a load balancer that will route the traffic to an alive node, re-get routing info and all will be well. What do you think?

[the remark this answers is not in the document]
mvanderkroon: ah I hadn't thought about the CANDIDATE stage, but I think you're right!

Your example of scenario 2 seems entirely on point, it's consistent with what we see happen every once in a while. 

I agree: connect afresh to the globally configure connection string and it should resolve. I see one possible risk: there might be a small window where the load balancer has not picked up the unavailability of node 1 and so there is a (very) small chance that connecting afresh might end up at node 1 - I'm not sure of the likelihood of this happening (it also depends on health check interval settings and cluster size), but it might be prudent to ensure some retry logic for the reconnect with a configurable MAX_RETRY_SECONDS (which should then be set by the user to a value greater than the health check interval, with a reasonable default - 10 seconds?).

neomodel/util.py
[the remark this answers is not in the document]
aanastasiou: Yes, this has [come up before](https://github.com/neo4j-contrib/neomodel/pull/385) too. Let's not worry too much about it at this point as there are a few pull requests that will have to be considered for their interactions as well when the time comes for a release.

neomodel/util.py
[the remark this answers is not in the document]
mvanderkroon: @robertlagrant is there any chance you can try this to see if it resolves the issues?

neomodel/exceptions.py
```diff
@@ -53,15 +56,21 @@ def __init__(self, db_node, current_node_class_registry):
         """
         self.db_node = db_node
         self.current_node_class_registry = current_node_class_registry
-        
+
     def __str__(self):
         node_labels = ",".join(self.db_node.labels)
-        ncr_items = list(map(lambda x:"{} --> {}".format(",".join(x[0]), x[1]),
-                             self.current_node_class_registry.items()))
+        ncr_items = list(
```
robertlagrant: This is just a style/preference thing, but it might be more readable to do the following:

```
ncr_items_multiline = "\n".join(
    "{} --> {}".format(",".join(node_labels), model_class)
    for node_labels, model_class in self.current_node_class_registry.items()
)
```

robertlagrant: This is just a style/preference thing, but it might be more readable to do the following:
```
ncr_items_multiline = "\n".join(
    "{} --> {}".format(",".join(node_labels), model_class)
    for node_labels, model_class in self.current_node_class_registry.items()
)
```

neomodel/util.py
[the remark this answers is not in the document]
mvanderkroon: @robertlagrant ping?

neomodel/exceptions.py
mvanderkroon: that definitely looks better, I use an autoformatter (black) that wreaked havoc on the existing formatting, hence the large amount of changes. 

Question is whether the implemented logic is solving your issues, can you check this?

robertlagrant: @mvanderkroon yes! I just came here to ask if it's ready for a test. I'll give it a go.

robertlagrant: I'm still getting a `NotALeaderError`. I can't see why, as I'm a bear of very little brain. I double-checked that the code I'm using is yours, and I can see the `except (ServiceUnavailableException, NotALeaderError, DatabaseUnavailableError` in your code, so that's very odd!

Here's the traceback:
```
2019-01-20 21:02:54.855 | INFO     | commands:do_second:80 - Attempting to write data to neo4j-2
2019-01-20 21:02:54.871 | ERROR    | commands:do_second:84 - Traceback (most recent call last):
  File "/commands.py", line 81, in do_second
    write_data()
  File "/usr/local/lib/python3.7/site-packages/neomodel/util.py", line 317, in wrapper
    return func(*args, **kwargs)
  File "/commands.py", line 56, in write_data
    return db.cypher_query(INSERT_QUERY)
  File "/usr/local/lib/python3.7/site-packages/neomodel/util.py", line 39, in wrapper
    return func(self, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/neomodel/util.py", line 261, in cypher_query
    raise exc_info[1].with_traceback(exc_info[2])
  File "/usr/local/lib/python3.7/site-packages/neomodel/util.py", line 233, in cypher_query
    results, meta = ([list(r.values()) for r in response], response.keys())
  File "/usr/local/lib/python3.7/site-packages/neomodel/util.py", line 233, in <listcomp>
    results, meta = ([list(r.values()) for r in response], response.keys())
  File "/usr/local/lib/python3.7/site-packages/neo4j/v1/api.py", line 736, in records
    self._session.fetch()
  File "/usr/local/lib/python3.7/site-packages/neo4j/v1/api.py", line 347, in fetch
    detail_count, _ = self._connection.fetch()
  File "/usr/local/lib/python3.7/site-packages/neo4j/bolt/connection.py", line 290, in fetch
    raise error
  File "/usr/local/lib/python3.7/site-packages/neo4j/bolt/connection.py", line 287, in fetch
    return self._fetch()
  File "/usr/local/lib/python3.7/site-packages/neo4j/bolt/connection.py", line 327, in _fetch
    response.on_failure(summary_metadata or {})
  File "/usr/local/lib/python3.7/site-packages/neo4j/v1/result.py", line 70, in on_failure
    raise CypherError.hydrate(**metadata)
neo4j.exceptions.NotALeaderError: No write operations are allowed directly on this database. Writes must pass through the leader. The role of this server is: FOLLOWER
```

aanastasiou: @robertlagrant, @mvanderkroon  There seem to be certain events that do not trigger a sync from the databsase server towards the driver. In the past, I have experienced this in relation to transactions where an operation was causing an error server-side, which however was only becoming known at client-side during a different transaction. This was confusing because it flagged errors at points that were not directly related to it.

Therefore, do you think that you could try [sync](https://neo4j.com/docs/api/python-driver/current/transactions.html#neo4j.neo4j.Session.Session.sync) on the session, at the point that is likely to be causing the problem? Perhaps this is one of those cases where the server has updated its state but this has not been propagated back to the client side (?). So, upon the first "fail", do a `sync` first and THEN retry the transaction. Hopefully this will update the client about which server has become the leader (?). Just an idea.

aanastasiou: @robertlagrant, @mvanderkroon Shall we include this change in an upcoming release? Did you guys try the `sync` before re-trying?

mvanderkroon: @aanastasiou I am busy these weeks with other activities, so I was not able to test it just yet. At the earliest I'd be able to look into it at the beginning of next week.

aanastasiou: @mvanderkroon Of course, I understand, there is no rush, if I have sometime from my side I will give it a try too. 

All the best

mvanderkroon: update from my end: we have been running into exactly the issue as @robertlagrant (countless `NOTALEADER` errors). However, in this case it seems unrelated to neomodel (or even the python-neo4j-driver): for us, causal clustering is just incredibly unstable. The causal cluster in fact gets stuck in a continuous LEADER election loop. I'm therefore quite certain that adding more `retry-logic` makes little sense on the application side. 

To me, 'causal clustering' is either fundamentally broken, or we are unlucky enough to encounter some sort of edge case bug. So far our experience with causal clustering has been far from ideal: unstable all the way through and write performance is ~6x worse than against unclustered. Our solution so far is to reduce the number of core nodes to 1 (essentially losing cluster capabilities) and keep our fingers crossed until we can migrate away from neo4j.

[fyi, seems related](https://github.com/neo4j/neo4j-python-driver/issues/276)

aanastasiou: @mvanderkroon Thanks for letting me know. @robertlagrant is there anything to add from your end on this (?).

aanastasiou:
Possibly [related](https://github.com/neo4j-contrib/neomodel/issues/303). There are certain conditions for which the client does not get updated about any error conditions immediately, unless the client does a `sync()`. This is done for performance reasons.

mariusconjeaud:
Can that be merged ? What would need to be done to merge this ? Or should it be closed ?

Issue #378, opened 2018-11-13
Upgrading to version 3.3.0 causes a KeyError on initialising connection to neo4j DB


jonadaly:
We have a neo4j instance, populated with the help of neomodel v3.2.9. The v3.3.0 of neomodel breaks when attempting to initialise the connection to the database:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/neomodel/util.py", line 150, in _object_resolution
    resolved_object = self._NODE_CLASS_REGISTRY[frozenset(a_result_attribute[1].labels)].inflate(
KeyError: frozenset({'Person'})
```

This appears to be as a result of the self._NODE_CLASS_REGISTRY changes included in v3.3.0, which appear to be breaking. Is this a known issue? Given that this is a minor version bump, we would not expect breaking changes to be included.

aanastasiou:
Hello @jdalydt 

Thanks for letting us know about this. No, it is not a known issue so far and I am sorry if it has caused a headache to you guys.

Can you please confirm if `import neomodel` comes **before** any other neomodel imports you might be doing on the file that raises the error?

This is related to the way neomodel creates a mapping between labels and native classes to be able to properly instantiate a class given a generic `Node` as retrieved from the database. To make sure that every call references the same `neomodel` "instance", `import model` has to be placed first.

All the best

jdalydt:
Adding `import neomodel` at the top of the file producing the error has no effect.

The full stack trace is:

```
File "./lib/python3.6/site-packages/neomodel/match.py", line 564, in get_or_none
    return self.get(**kwargs)
  File "./lib/python3.6/site-packages/neomodel/match.py", line 548, in get
    result = self._get(limit=2, **kwargs)
  File "./lib/python3.6/site-packages/neomodel/match.py", line 539, in _get
    return self.query_cls(self).build_ast()._execute()
  File "./lib/python3.6/site-packages/neomodel/match.py", line 445, in _execute
    results, _ = db.cypher_query(query, self._query_params, resolve_objects=True)
  File "./lib/python3.6/site-packages/neomodel/util.py", line 32, in wrapper
    return func(self, *args, **kwargs)
  File "./lib/python3.6/site-packages/neomodel/util.py", line 202, in cypher_query
    results = self._object_resolution(results)
  File "./lib/python3.6/site-packages/neomodel/util.py", line 162, in _object_resolution
[...]

PR #402, opened 2019-01-24
#378, converted thread-local `Database` instance to single Database instance

jberends:
See discussiong on #378 

* fixed re-initalisation of the `Database` object in each new thread.
* tested with django using the repo: https://github.com/jberends/django_neo_3.3.0/tree/master


aanastasiou:
Covered by #403, thank you very much.

PR #403, opened 2019-01-25
Refactor of Node Class Registry container used by util.Database.

oikonomou:
Issue #378 describes the problem in neomodel v3.3.0 in which the new node-class-registry container fails to map node class names to node class instances. 
The problem was narrowed to the place that container is initialized which is during initialization of Database instance.
The refactor in this pull request is the creation of a singleton class (NodeClassRegistry) whose subclasses share the same container for the aforementioned mapper. Therefore, on every Database initialization, the container is intact and not initialized for multiple times.
The Database class inherits from this new class.

### DIFF
--- a/neomodel/util.py
+++ b/neomodel/util.py
@@ -42,7 +42,22 @@ def clear_neo4j_database(db):
     db.cypher_query("MATCH (a) DETACH DELETE a")
 
 
-class Database(local):
+class NodeClassRegistry:
+    """
+    A singleton class via which all instances share the same Node Class Registry.
+    """
+    # Maintains a lookup directory that is used by cypher_query
+    # to infer which class to instantiate by examining the labels of the
+    # node in the resultset.
+    # _NODE_CLASS_REGISTRY is populated automatically by the constructor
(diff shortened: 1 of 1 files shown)


aanastasiou:
@oikonomou Thank you very much for your effort with this. Now, I don't know if this is work in progress but I thought I would note at this point that simply introducing this kind of hierarchy is not going to work as intended.

The metaclass form described [here](https://stackoverflow.com/questions/6760685/creating-a-singleton-in-python) would work better in this case I think.

In other words, a new type is created that maintains an internal dictionary of instances it has created to make sure that only one is shared within a specific context. That `Singleton` "bit" can then give birth to either `Database` or `NodeClassRegistry` (as per your commit) in an attempt to keep a single reference point for the dictionary.

But even if we do this, we have to ensure that:

1. The mappings of two entirely different applications do not get blended together
2. If it is impossible to separate mappings from two different applications, then an exception should be thrown at [this point](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/core.py#L174).

All the best

Issue #404, opened 2019-01-27
Feature request: add lazy option to QueryBuilder


machosec:
Hi, 
The `StructuredNode` class create/merge (`create_or_update`) functions has the `lazy` argument to only retrieve the id of the node tied to neo4j instead of all of its properties. This feature is great for many use cases.
I didn't find such option within the `QueryBuilder` that is used by the `NodeSet` and `Traversal` classes which seemed trivial to implement. Is there any reason why this feature does not exist? I guess this could be implemented in the `build_node` function. I can make a PR with this feature if you like.

Keep up the good work!

machosec:
Hi @MardanovTimur 
The reason to make it `lazy` is to avoid returning the entire neo4j node from the beginning. Retrieving lots of nodes with non-indexed properties is expensive. a pythonic solution like `partial` does not solve this issue unless I don't understand the usage

aanastasiou:
@machosec If I understand this correctly, you want the **last** call to only return IDs, rather than any intermediate ones because these intermediate might need the attributes for filtering (?), is that right?

PR #406, opened 2019-01-28
add lazy option to NodeSet get() and all()

machosec:
Following #404 I added a "`lazy`" switch to `NodeSet` `get()` and `all()` to allow returning nodes IDs instead of the full node properties.
The code should not break any previous functionality/behavior as the logic stays the same without explicitly specifying the optional argument.
I decided to put the return logic in the `_exeucte` function to keep the code style in `_constains` and `_count` which also change the return (`self._ast['return']`) behavior

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -440,7 +440,10 @@ class QueryBuilder(object):
         self._query_params[place_holder] = node_id
         return self._count() >= 1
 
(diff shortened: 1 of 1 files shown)


aanastasiou:
This is great thank you very much, will have a look and merge.

PR #409, opened 2019-02-07
Added safeguard for the accidental redefinition of a class mapping

aanastasiou:
Please see `doc/extending.rst` section "Caveats" for a detailed explanation of the safeguard added with this commit.

An exception (`ClassAlreadyDefined`) is raised, if a class is attempted to be defined that happens to have exactly the same set of labels with another class that may have already been defined within the node-to-class registry.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,6 +23,13 @@ An Object Graph Mapper (OGM) for the neo4j_ graph database, built on the awesome
     :scale: 100%
     :target: https://neomodel.readthedocs.io/en/latest/?badge=latest
 
+
+! Support for Python 2.7 !
+==========================
+**Please note:** Version 3.3.1 will be the last version that `neomodel` provides suppport for Python 2.7. By its
+next major release, the project will focus solely on Python 3.*.
+
+
 Documentation
 =============
 
@@ -33,7 +40,7 @@ Available on readthedocs_.
 Requirements
 ============
 
-- Python 2.7, 3.4+
+- Python 2.7 (Up to version 3.3.1), 3.4+
 - neo4j 3.0, 3.1, 3.2, 3.3
 
 Installation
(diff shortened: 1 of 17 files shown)


aanastasiou:
@oikonomou, @jberends, @mostafa Can I please ask you to cast your eyes over the changes in this commit and let me know your feedback?

mostafa:
@aanastasiou Thank you for the ticket and fix.

I've tested your branch and it works fine, meaning it raises the exception on application start which is way better than figuring it out while you are running the app.

Also I have changed my tests to reflect the new changes.

aanastasiou:
@mostafa You are welcome. 

Can you talk a little bit more about how would that work?

* The node-to-class registry maintains a mapping of `{label}->{class}`.
* When a `neo4j::Node` is retrieved from the database, its set of labels is used to infer which native application specific class should be instantiated.
* The mapping is implied to be unique. Therefore, it is impossible to resolve between two mappings that happen to have the same labels but could well resolve to different classes (from different namespaces for example). Maybe it is possible to do this in Python via its own namespace management, but that is not carried over to the way `neomodel` works.
* Perhaps we can extend its abilities to infer which class to instantiate but that would require more data than simply the set of labels. 

That might be a corner case anyway, so we raise an exception to warn against a problematic situation and also describe the condition in the documentation.

I am slightly worried that anything more complicated than that (a straightforward mapping) might introduce delays but happy to see alternatives.

mostafa:
The reason for mentioning this is that I tried to run my tests with pytest by using its pytest-xdist (concurrency) and pytest-forked (parallelism) extensions, but none of them helped alleviate the issue of duplication, so I renamed the classes and everything worked, as expected.

Then I thought that maybe we could use something like `{(labels, (properties)): class}` mapping (a tuple to class mapping basically) to differentiate between the same labels but different parameters. That would also be a corner-case, but it would help with concurrency and parallelism in code, where you might have introduced multiple labels in different modules with different set of properties in each of them.

We can also leave it as it and do not scrutinize more, since it is a corner-case.

aanastasiou:
@mostafa Yes, this came up over here too but it makes you wonder where do you stop because the behaviour depends very much on context as well.

I think that where we are is a good compromise. I have a little bit more about why a use-case such as yours seems to fail in the documentation but it is written from the point of view of dealing with Neo4J clusters. I might add a thing or two about writing tests as well, in the same section, to at least warn against this kind of trouble.

All the best

jberends:
@aanastasiou 
Passed the point of no-return, but wanted to say that I concur with your commit.

PR #410, opened 2019-02-12
Fix log format for logging cypher queries and response time

lalitpatel:
use `{:f}` instead of `{:.2g}`


lalitpatel:
Example:
```python
import time
start = time.time()
time.sleep(10.5)
end = time.time()
print "{:.2g}".format(end-start)  # 11
print "{:f}".format(end-start)  # 10.504036
```

lalitpatel:
Closing this PR since it has been open for a long time without a review.

PR #412, opened 2019-02-13
Fix string representation of exception ModelDefinitionMismatch

lalitpatel:
For neo4j-driver 1.6.x Node.labels is a `@property` and not a method

https://github.com/neo4j/neo4j-python-driver/blob/1.6/neo4j/v1/types/graph.py#L206


aanastasiou:
Thank you, there are quite a few changes like this that will be scheduled for the next major release.

lalitpatel:
Closing this PR since it has been probably addressed in the next release.

Issue #259, opened 2017-06-18
neomodel closes sessions only after object is deleted


liranbg:
Using neomodel 3.2.5

I figured when using transaction with db.begin() and db.commit() it does not close the session immediately but when the object itself is being destroyed.

I used the following test to determine when the session is being close by using logging's output to check when the object is being deleted (before the tearDown() or after)
in test_session_closing_neomodel -> it closes session only after deleted
in test_session_closing_neodriver -> it closes session when transaction is ended

```python
import unittest
import logging
from neomodel import db
from neo4j.v1 import GraphDatabase
logging.basicConfig(level=logging.DEBUG)


class NeoTest(unittest.TestCase):
    def setUp(self):
        pass

    def tearDown(self):
        pass

    def test_session_closing_neomodel(self):
        db.set_connection("bolt://neo4j:neo4j@localhost:7687")
        db.begin()
        db.cypher_query("MATCH (n) RETURN n")
        db.commit()

    def test_session_closing_neodriver(self):
        driver = GraphDatabase.driver("bolt://localhost:7687", auth=("neo4j", "neo4j"))
        with driver.session() as session:
            with session.begin_transaction() as tx:
                tx.run("MATCH (n) RETURN n")
                tx.commit()
```

robinedwards:
Thanks for raising this issue. Coupled to it is also the issue with the connection clean up causing some strange exceptions in neo4j_drivers destructor. Spent a while trying to figure it out to no avail but will let you know as soon as I find a solution

liranbg:
@robinedwards Thanks a lot. let me know if there's any help needed

aanastasiou:
@liranbg Would this still be of interest? Looks like adding an explicit `close` [here](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/util.py#L125) (?)

liranbg:
@aanastasiou - That indeed would be more declarative rather than let python's gc do it

aanastasiou:
@liranbg Thanks for such a quick response. Just to understand the severity of the problem, what is the worst that can happen from leaving this up to the system? Besides the exceptions on exit (which I have experienced myself as well) do you think this could be causing any transaction related problems too? (for example).

liranbg:
sorry for this late response.
I believe, that for a huge transaction loop, the behavior expectation would be freeing the allocated fd once an iteration is done. and since it is being dangled, we might hit the maximum fd limit

aanastasiou:
[...]

PR #415, opened 2019-02-18
Upgrade to Neo4J 3.5 and Neo4j Python Driver 1.7.1

devinbarry:
I don't expect this to be merged, but I am putting it here in case anybody finds it useful.

Upgraded to the latest versions of Neo4J db and Python driver. All tests pass except for two to do with spacial stuff. I don't care about any of the spacial stuff so I am not going to dig deep to find out what is wrong with those tests. They're just skipped for now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,16 @@
 language: python
 sudo: required
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
+  - "3.7"
 env:
-  - NEO4J_VERSION="3.0.10"
-  - NEO4J_VERSION="3.1.7"
-  - NEO4J_VERSION="3.2.9"
-  - NEO4J_VERSION="3.3.3"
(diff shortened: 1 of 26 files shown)


devinbarry:
Spent ages trying to fix Travis. Finally got it building for Python 3.4 - 3.7. I am running my tests locally on Neo4j 3.5.3 but on Travis when using 3.5.3 one of the tests fails with an SSL error across all Python versions. Pretty interesting stuff. No idea what it means

devinbarry:
I thought that maybe using the context manager for a session (and thus ensuring connections are properly closed) might fix the SSL errors I am getting on Travis. I am using session now exactly as the driver docs say, but it still doesn't fix Travis.

aanastasiou:
@devinbarry Do you mean something like [this](https://github.com/neo4j-contrib/neomodel/issues/259) (?)

Issue #417, opened 2019-02-20
nodes.get(location=NeomodelPoint((0, 0), crs='cartesian') does not work


DmitriyG228:
Hi, getting nodes with location, specified with NeomodelPoint give an error "Invalid instantiation via no arguments", while setting  node in the same way works well (below it gives ConstraintError). What is the right way of getting nodes with specific location?

```
location=(51.3454, -6.2434)
try:
    property = Property.nodes.get(location=NeomodelPoint(location,crs='cartesian'))
except:
    property = neo4j.Property(location=NeomodelPoint(location,crs='cartesian')).save()
```

ValueError: Invalid instantiation via no arguments. A Point needs default values either in x,y,z or longitude, latitude, height coordinates

ValueError: Invalid instantiation via no arguments. A Point needs default values either in x,y,z or longitude, latitude, height coordinates

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/dmitriy/anaconda3/lib/python3.7/site-packages/neomodel/util.py", line 211, in cypher_query
    response = session.run(query, params)
  File "/Users/dmitriy/anaconda3/lib/python3.7/site-packages/neo4j/v1/api.py", line 331, in run
    self._connection.fetch()
  File "/Users/dmitriy/anaconda3/lib/python3.7/site-packages/neo4j/bolt/connection.py", line 287, in fetch
    return self._fetch()
  File "/Users/dmitriy/anaconda3/lib/python3.7/site-packages/neo4j/bolt/connection.py", line 327, in _fetch
    response.on_failure(summary_metadata or {})
  File "/Users/dmitriy/anaconda3/lib/python3.7/site-packages/neo4j/v1/result.py", line 70, in on_failure
    raise CypherError.hydrate(**metadata)
neo4j.exceptions.ConstraintError: Node(2484) already exists with label `Property` and property `location` = {geometry: {type: "Point", coordinates: [51.3454, -6.2434], crs: {type: link, properties: {href: "http://spatialreference.org/ref/sr-org/7203/", code: 7203}}}}

aanastasiou:
@DmitriyG228 Can you please try `.get`ting with `NeomodelPoint(x=location[0], y=location[1], crs='cartesian')`?

DmitriyG228:
[...]

PR #418, opened 2019-02-21
Fixed NeomodelPoint instantiation via copy, deepcopy operations.

aanastasiou:
Bug fix in response to issue #417.


aanastasiou:
Just as a note: When I run the tests locally, I get 1 failure on the transactions test which I am pretty sure that is because of my setup. The rest of the tests and specifically those about the spatial data types pass without any major complaints.

I have not looked any further into these specific failures reported here, it might be something to do with configuration.

Issue #422, opened 2019-03-06
BUG: Object Resolution raises ModelDefintionMismatch on nodes with extra labels


henryfjordan:
We have some node classes that we use extra labels to describe/query. Because of the new Object Resolution feature, querying for these nodes now breaks our application.

A good example would be an `archived` label. We have some Node that we might want to soft-delete by adding the `:Archived` label then doing a `MATCH (n) WHERE NOT n:Archived` to get still valid nodes. We use labels because they are more performant than properties for this task.

The problem for us lies in the following code in `_object_resolution()`:

```python
try:
    # Primitive types should remain primitive types, 
    #  Nodes to be resolved to native objects
    resolved_object = a_result_attribute[1]
    
    if type(a_result_attribute[1]) is Node:
        resolved_object = self._NODE_CLASS_REGISTRY[frozenset(a_result_attribute[1].labels)].inflate(
            a_result_attribute[1])
        
    if type(a_result_attribute[1]) is list:
        resolved_object = self._object_resolution([a_result_attribute[1]])                    
    
    result_list[a_result_item[0]][a_result_attribute[0]] = resolved_object
    
except KeyError:
    # Not being able to match the label set of a node with a known object results 
    # in a KeyError in the internal dictionary used for resolution. If it is impossible 
    # to match, then raise an exception with more details about the error.
    raise ModelDefinitionMismatch(a_result_attribute[1], self._NODE_CLASS_REGISTRY)
```
https://github.com/neo4j-contrib/neomodel/blob/8d02b4a0d15a980ec2ac2653846fd01727384638/neomodel/util.py#L165

This code should be checking for the best match, not an exact match. If a node has 7 Labels, including all 5 required for a certain class, the presence of 2 extra labels shouldn't break our queries. I'm not sure exactly how I would implement this in a performant way, but I know raising an error is very limiting in the use-cases this library supports.

Alternatively, we might be able to build a class to describe every single combination of Labels that is possible in our application, but I consider the need to do that in order to upgrade a breaking change and should have triggered a major version change on this library.

Please let me know how I can help resolve this issue without breaking our use-case.

henryfjordan:
Here is a branch with what I think is a fix

https://github.com/neo4j-contrib/neomodel/compare/master...henryfjordan:fix-object-resolution

aanastasiou:
@henryfjordan I can see the inconvenience but at the same time we are going to have to try and balance a few things here.

[...]

PR #425, opened 2019-03-12
Optional labels

henryfjordan:
Adds optional labels, then registers classes to the map with these in mind.

One test is included.

You just add some `__optional_labels__` to a class like so:
```python
class Person(StructuredNode):
    __optional_labels__ = ["Alive", "Contributor"]
    name = StringProperty()
```

This solution has to examine the optional labels only once per class on startup and doesn't affect query time performance. It just packs the class dict a little tighter but lookup is O(1) there.

I took a swing at doing the label diff calculations at query time, but I really feel like this is a better solution. To determine all this at query time, you would have to do something like the following:

```diff
diff --git a/neomodel/util.py b/neomodel/util.py
index b1946a8..664f6b6 100644
--- a/neomodel/util.py
+++ b/neomodel/util.py
@@ -162,8 +162,18 @@ class Database(local, NodeClassRegistry):
                     resolved_object = a_result_attribute[1]

                     if type(a_result_attribute[1]) is Node:
-                        resolved_object = self._NODE_CLASS_REGISTRY[frozenset(a_result_attribute[1].labels)].inflate(
-                            a_result_attribute[1])
+                        existing_labels = set(a_result_attribute[1].labels)
+
+                        important_labels = frozenset(existing_labels.difference(global_optional_labels))
+                        optional_labels = existing_labels.intersection(global_optional_label_list)
+
+                        matching_class = self._NODE_CLASS_REGISTRY[important_labels]
+
+                        if not optional_labels.issubset(set(matching_class.__optional_labels__)):
+                            # There must be unaccounted for labels present
+                            raise ModelDefinitionMismatch(a_result_attribute[1], self._NODE_CLASS_REGISTRY)
+
+                        resolved_object = matching_class.inflate(a_result_attribute[1])

                     if type(a_result_attribute[1]) is list:
                         resolved_object = self._object_resolution([a_result_attribute[1]])
```
Note that this approach would require either maintaining a list `global_optional_labels`, effectively detaching them from the models, which I didn't like. You could gather the list from all the models, but that is inflexible (and requires recompilation all the time).

Anyway, let me know what you think!

### DIFF
--- a/doc/source/extending.rst
+++ b/doc/source/extending.rst
@@ -21,6 +21,26 @@ Creating purely abstract classes is achieved using the `__abstract_node__` prope
             self.balance = self.balance + int(amount)
             self.save()
 
+
+Optional Labels
+---------------
+Sometimes it is useful to allow a node to have extra labels in addition to the
+ones Neomodel defines by default using class names.
+
+Nemodel constructs sets of labels to give to a node in Neo4j by looking at the
+names of the node classes that node is/inherits from. It also constructs a
+mapping of the reverse, expected node labels to node classes, in order to do
+object resolution.
+
+In order for object resolution to work on Node Classes that could have extra
+labels, the `__optional_labels__` property must be defined as a list of strings::
+
+    class Shopper(StructuredNode):
+        __optional_labels__ = ["SuperSaver", "SeniorDiscount"]
+        balance = IntegerProperty(index=True)
+
+.. warning:: The size of the node class mapping grows exponentially with optional labels. Use with some caution.
+
 Mixins
 ------
 Mixins can be used to share functionality between nodes classes::
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -183,8 +183,9 @@ class NodeMeta(type):
                 for i in range(1, len(optional_label_set) + 1)
                 for x in combinations(optional_label_set, i)
             ]
+            possible_label_combinations.append(base_label_set)
 
-            for label_set in [base_label_set, *possible_label_combinations]:
+            for label_set in possible_label_combinations:
                 if label_set not in db._NODE_CLASS_REGISTRY:
                     db._NODE_CLASS_REGISTRY[label_set] = cls
                 else:
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -175,7 +175,7 @@ class NodeMeta(type):
                 install_labels(cls)
 
             base_label_set = frozenset(cls.inherited_labels())
-            optional_label_set = set(cls.optional_labels())
+            optional_label_set = set(cls.inherited_optional_labels())
 
             # Construct all possible combinations of labels + optional labels
             possible_label_combinations = [
@@ -487,7 +487,7 @@ class StructuredNode(NodeBase):
                 scls, '__abstract_node__')]
 
     @classmethod
-    def optional_labels(cls):
+    def inherited_optional_labels(cls):
         """
         Return list of optional labels from nodes class hierarchy.
 
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -169,7 +169,7 @@ class NodeMeta(type):
             )
 
             cls.__label__ = namespace.get('__label__', name)
-            cls.__optional_labels__ = namespace.get('__optional_labels__', name)
+            cls.__optional_labels__ = namespace.get('__optional_labels__', [])
 
             if config.AUTO_INSTALL_LABELS:
                 install_labels(cls)
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -1,6 +1,7 @@
 import re
 import sys
 import warnings
+from itertools import combinations
 
 from neomodel import config
 from neomodel.exceptions import DoesNotExist, ClassAlreadyDefined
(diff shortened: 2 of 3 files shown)


henryfjordan:
Fixes #422

henryfjordan:
Closing a re-opening so travis will rerun

aanastasiou:
Hi

A couple of notes in no particular order:

1. As an overall remark, my understanding of this is that given two sets of labels `A=['LABEL_1', 'LABEL_2'], B=['OL_1', 'OL_2']` for some `SomeClass`, this initialisation populates the `_NODE_CLASS_REGISTRY` with:
    ```
    LABEL_1, LABEL_2, OL_1 --> SomeClass
    LABEL_1, LABEL_2, OL_2 --> SomeClass
    LABEL_1, LABEL_2, OL_1, OL_2 --> SomeClass
    LABEL_1, LABEL_2 --> SomeClass
    ```

    1. How are derived classes supposed to deal with `__optional_labels__`? 
        1. Are they supposed to inherit them?   
            * Inheriting the labels means that even if the derived class does not specify any new `__optional_labels__` it **could** appear with any of the optional labels of its mother class.
        2. Or is it that `__optional_labels__` belong to the class they are defined in and that's that?

    2. As a side note: `_NODE_CLASS_REGISTRY` receives `2**n` entries pointing to the same class, where `n` is the length of `__optional_labels__` and `**` denotes exponentiation (The `n=0` case produces the current behaviour) . With `n=3` the `_NODE_CLASS_REGISTRY` grows by 8 entries that are pointing to the same class.
        * If a derived class "inherits" labels from further up its lineage, then the amount of new entries *per class* becomes `k * 2^M` where `M = n_0 + n_1 + n_2 + ...`, the n-subscript is interpreted as *optional labels of a class, `subscript` levels up the hierarchy* and `k` is the number of classes derived from a given mother class at the same level. So `n_0` is the number of optional labels of the class itself, `n_1` are the optional labels of the mother class, `n_2` is the grand-mother and so on. 
        * Example 1:
            ```
            class Base(StructuredNode):
                pass

            class Alpha(Base):
                __optional_labels__ = ["TIK", "TOK"]
            
            #k=1
            class Beta(Alpha):
                pass
            ```
            This produces 1 entry for `Base`, 4 for `Alpha` and another 4 for `Beta`.

        * Example 2:
            ```
           class Base(StructuredNode):
                pass

            class Alpha(Base):
                __optional_labels__ = ["TIK", "TOK"]
            
            #k=1
            class Beta(Alpha):
                pass

            #k=2
            class Gamma(Alpha):
                pass

            #k=3
            class Delta(Alpha):
                pass 
            ```
           This is the most common pattern when you are trying to offer alternative options for entities in a data model. This produces `3 * 2**2 == 12` entries + 1 for `Base` = 14 labels.

        * Example 3:
            ```
           class Base(StructuredNode):
                pass

            class Alpha(Base):
                __optional_labels__ = ["TIK", "TOK"]
            
            class Beta(Alpha):
                pass

            class Gamma(Beta):
               __optional_labels__ = ["FLIP", "FLOP"]

            class Delta(Gamma):
                pass 
            ```
           This produces 1 for `Base`, 4 for `Alpha`, 4 for `Beta`, 16 for Gamma, 16 for `Delta` for a total of 41.

3. Obviously, if the model gets extended with a class that happens to have a name that matches one of the `__optional_labels__` defined earlier then this would lead to a `ClassAlreadyDefined` exception. I wonder if it would be possible to modify the message at that point to give a hint that this could have been caused because of an optional label definition (the reverse is already caught) (?).

All the best

henryfjordan:
Thanks for taking a look!

To answer 1i, the classes are supposed to inherit `__optional_labels__`, accomplished through the `optional_labels()` function which loops through `cls.mro()` in a very similar fashion to `labels()`. Personally I don't care whether they inherit the optional labels or not but it felt like the more complete option.

As to point 1ii, yeah I agree that the combinatorics can get ugly. I figure that in practice the map will not get huge but perhaps we should add a `log.warn()` if the class map grows unreasonably large (1000 entries?). That being said, the only cost of a larger class map is more RAM consumption, so I don't see it as that big of a deal. For a little reference, our particular use case has 4 optional labels.

Lastly, to point 2, I'll take a look. I think that's a good idea. It should be possible to check the `label_set` you have vs the labels on the class and modify the text accordingly

aanastasiou:
> To answer 1i, the classes are supposed to inherit `__optional_labels__`,...

I suspected they might and this mixing of concepts where a label plays a dual role, both as a "tag" and as a "type" for nodes would be my biggest reservation about this feature.

> As to point 1ii, yeah I agree that the combinatorics can get ugly. I figure that in practice the map will not get huge...

There is always room for abuse of a concept if it appears to be a convenient shortcut.

> ...but perhaps we should add a log.warn() if the class map grows unreasonably large (1000 entries?). That being said, the only cost of a larger class map is more RAM consumption, so I don't see it as that big of a deal. 

There is definitely more memory usage. But also, as the experience from fixing #378 showed, WSGI applications spawn a new process to serve a request. Within the context of a new process, `neomodel` gets re-initialised, which basically means re-importing the modules that define the application's data model. So, inevitably, the combinatorics part (plus the memory operations) would insert some delay per request. 

Of course, for a small number of labels, this might not appear to be prohibitive. But by way of inheritance, it might be easy to get into a problematic region.

> For a little reference, our particular use case has 4 optional labels...

That's 16 entries pointing to the same class. I did not think your use-case would have more than 2-3 optional labels.

> Lastly, I didn't modify any documentation. Should I? 

Which is where I am getting at with this post, rather than being entirely negative with this feature. Now, personally I do not think that it's a good idea because a) It mixes concepts and b) It can get "ugly" quickly.  **BUT**, as a feature, it might be useful if one knows what they are doing *or* if one knows in advance exactly what they are getting into. 

So, yes, please add a little bit more in the documentation about this feature and let's have a look about it again later.

Also, I am going to try and check with others in Neo4J regarding the performance aspect of this. The reason I am thinking about doing this is because from the point of view of indexing, there is absolutely no difference in indexing a categorical property and indexing labels. Quite simply, a Label is an inherent categorical property of a Node. So, it would kind of bug me if they are not already (re)using the same indexing algorithm...this is why I would like to ask. And also, if we realise that there is no difference later on, this feature might be marked for progressive retirement over the better practice of establishing a property.

All the best

aanastasiou:
@henryfjordan Please see [here](https://community.neo4j.com/t/performance-of-node-labels-vs-attributes/6107/3)

PR #428, opened 2019-03-19
Add a DateTimeFormatProperty which allows integrating with custom date format

ysn2233:
Add a DateTimeFormatProperty so user can create a time property like that.
``` python
# In neo4j the property is like YYYY-mm-dd HH:MM:SS
some_format = "%Y-%m-%d %H:%M:%S"
timestamp = DateTimeFormatProperty(format=some_format)
```
Also add a test in test_properties.py to test the function.

### DIFF
--- a/doc/source/properties.rst
+++ b/doc/source/properties.rst
@@ -4,14 +4,15 @@ Property types
 
 The following properties are available on nodes and relationships:
 
-==============================================  ===========================================================
-:class:`~neomodel.properties.AliasProperty`     :class:`~neomodel.properties.IntegerProperty`
-:class:`~neomodel.properties.ArrayProperty`     :class:`~neomodel.properties.JSONProperty`
-:class:`~neomodel.properties.BooleanProperty`   :class:`~neomodel.properties.RegexProperty`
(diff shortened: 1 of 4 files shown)


aanastasiou:
@ysn2233 Thanks, but why not implement this as an option to the existing `Date` and `DateTime` properties?

henryfjordan:
👍 for making this a new property type. This makes slowly migrating to the new type easier

ysn2233:
> @ysn2233 Thanks, but why not implement this as an option to the existing `Date` and `DateTime` properties?
Thank you for relying.
The reason why I create a new property is that:
1. In current implementation DateProperty accepts a date instance rather than datetime. 
2. In current implementation DateTimeProperty it transfers a datetime instance to unix timestamp.
Therefore I think if I integrate the feature into these two properties without break current design, it might be vague for users.

aanastasiou:
@ysn2233 Sounds good. Do you think you can maybe added a few words about the property [here](https://neomodel.readthedocs.io/en/latest/properties.html)?

ysn2233:
@aanastasiou I have add some description in doc/source/properties.rst

Issue #438, opened 2019-04-22
New function required: StructuredNode.merge


John-Theo:
Awesome awesome work!

I want to do `a = Article(title='some title').merge()` **like actual MERGE action** in Cypher.
The lines below are from `neomodel.core`. The `if` statement on line 520 decide whether to update or create the node. **THE PROBLEM IS**: How should the node has the attribute of `id` when it was just created?
OR, **is there already a way** to do something like `a = Article(title='some title').merge()`? I went through the documentation and the codes, found `get_or_create` and `create_or_update`, but they are not what I want.

Many thanks!

https://github.com/neo4j-contrib/neomodel/blob/cca5de4c4e90998293558b871b1b529095c91a38/neomodel/core.py#L512-L534

aanastasiou:
> Awesome awesome work!

@John-Theo On behalf of the ~30 developers who have contributed to `neomodel`, I am just going to say "Thanks" :)

> How should the node has the attribute of id when it was just created?

It depends on what are you trying to do but from what you describe, the best thing to do is to use `get_or_create` and create `Article` entities en masse.
The `if` statement you are pointing out is a design choice. If the node has an `id` attribute set, then it means that it is saved already (i.e. a `Node` exists in the database server). If you are trying to update the node then once you have a handle to it (e.g. `my_node = Article.nodes.get(some_unique_attribute="some_value")`), change the attribute values and do a `my_node.save()` again.

Hope this helps.

John-Theo:
Sorry for not describing it clearly. 

> It depends on what are you trying to do but from what you describe

The `MERGE` function I mentioned can be found [[here]](https://neo4j.com/docs/cypher-manual/current/clauses/merge/). It allows to create or update(depends on unique indexes) a node within ONE cypher query.
I have written a few lines(it works nicely in my own environment) and made a **pull request** #440  (it is my first time to contribute to an open source project TAT, really hope it turns out well). Somehow **the Travis build failed**. 
Will you be so kind as to maybe take a look at it? @aanastasiou 

Many thanks in advance.

aanastasiou:
@John-Theo What is it that you are trying to achieve? Let's not worry about how it is going to be done at the moment.

John-Theo:
[...]

PR #440, opened 2019-04-23
Add merge functionality

John-Theo:
Completed the functionality described in #438 .
This StructureNode.merge() function allows users to create a node without worrying whether it is already in the graph or not.

e.g.

```python
class Article(StructuredNode):
    title = StringProperty(unique_index=True, required=True)
    param = StringProperty()
```
```python
article = Article(title='some title', param='value').merge()
```
This line is translate into Cypher as:
```Cypher
MERGE (a:Article {title: 'some title'})
SET a.param = 'value'
RETURN a
```
PLEASE NOTE: One and ONLY one unique index required.


John-Theo:
Things seems right with NEO4J_VERSION="3.4.0"...
All I've done is adding a function to core.StructureNode. Works fine on my computer. So confused.

PR #445, opened 2019-05-03
Added `max_length` constraint on `StringProperty`

aanastasiou:
Based on initial work contributed by @Lazy-Y.


aanastasiou:
Seems that recent changes to `neo4j-driver-1.7.2` will have to trigger a fix on the recently introduced spatial data types part, as the storage retrieval test returns "cartesian points not supported":/.

PR #456, opened 2019-06-14
Correct inconsistent naming kls to cls

elena:
Pretty sure that `cls` is universally used, `kls` doesn't seem to be used anywhere else in this library.

### DIFF
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -114,8 +114,8 @@ def install_all_labels(stdout=None):
     if not stdout:
         stdout = sys.stdout
 
-    def subsub(kls):  # recursively return all subclasses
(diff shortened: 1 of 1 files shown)


aanastasiou:
Ah, good catch, thanks. I wonder if it was a slip of the keys for one of our Germanic contributors.

elena:
Wouldn't dare diss the use of 'kls', nor the quality of the fine and deeply respectable contributors for whom this is languages appropriate (actually I thought it came from language Java or something like this). But the stupid inconsistency was bothering me and then it occurred to me that *I* could actually tidy it ... this is the spirit of FOSS for sure ^_^

PR #458, opened 2019-07-12
Teach NodeSet to filter relationships against other NodeSets

eyalgr:
NodeSet.has can accept a NodeSet value and filter the relationship
against it, like so:
Person.nodes.has(pet=Pet.nodes.filter(name='kitty'))


eyalgr:
The implementation is broken in the general case:
Person.nodes.has(child=Person.nodes.has(child=Person.nodes.filter(name='john')))

Issue #284, opened 2017-11-14
ArrayProperty does not support kind of "contains" filter


slava-poddubsky:
Model example:
```python
from neomodel.core import StructuredNode
class User(StructuredNode):
    permissions = ArrayProperty(StringProperty(choices=['view', 'add', 'edit', 'delete']), required=True)

User.nodes.filter(permissions__icontains=['add', 'edit'])
```

Any ideas?

adamdelman:
This should probably work something like this:

```MATCH (n) 
WHERE single(x IN n.permissions WHERE x IN ['add', 'edit'])
RETURN n;
```


or

```
MATCH (n) 
WHERE any(x IN n.permissions WHERE x IN ['add', 'edit'])
RETURN n;
```

greydot:
Any update on this?

mariusconjeaud:
Check #820

PR #460, opened 2019-07-13
Feature: NodeSet.has filtering against a NodeSet

eyalgr:
NodeSet.has can accept a NodeSet value and filter the relationship against it, like so:
``` python
Person.nodes.has(pet=Pet.nodes.filter(name='kitty'))
```
NodeSet.has can accept a NodeSet that itself uses a NodeSet relationship filtering:
``` python
Person.nodes.has(
    child=Person.nodes.has(
        child=Person.nodes.filter(name='john')
    )
)
```

Node naming convention in QueryBuilder was modified - underscores in node names are duplicated as an escaping mechanism.

### DIFF
--- a/doc/source/queries.rst
+++ b/doc/source/queries.rst
@@ -128,6 +128,14 @@ The `has` method checks for existence of (one or more) relationships, in this ca
 
 This can be negated by setting `suppliers=False`, to find `Coffee` nodes without `suppliers`.
 
+Furthermore, `has` can check for a relationship to a specific set of nodes. For example, it is possible to find coffee
+that can be delivered for cheap, by passing a suppliers `NodeSet`::
+
+    Coffee.nodes.has(suppliers=Supplier.nodes.filter(delivery_cost__lt=2))
+
+The `has` function doesn't accept native python collections, only `NodeSet` or Boolean values. This is important
+because `all`, `get` and slicing (see following section) return a native collection and not a `NodeSet`.
+
 Iteration, slicing and more
 ===========================
 
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -200,7 +200,10 @@ def process_has_args(cls, kwargs):
         elif value is False:
             dont_match[rhs_ident] = rel_definitions[key].definition
         elif isinstance(value, NodeSet):
-            raise NotImplementedError("Not implemented yet")
+            if value.source_class is not rel_definitions[key].definition['node_class']:
+                raise ValueError('Relationship "{0}" is not of the same type as the provided NodeSet.'.format(key))
+            match[rhs_ident] = rel_definitions[key].definition.copy()
+            match[rhs_ident]['node_set'] = value
         else:
             raise ValueError("Expecting True / False / NodeSet got: " + repr(value))
 
@@ -215,8 +218,8 @@ class QueryBuilder(object):
         self._place_holder_registry = {}
         self._ident_count = 0
 
-    def build_ast(self):
-        self.build_source(self.node_set)
+    def build_ast(self, alt_node_name=None):
+        self.build_source(self.node_set, alt_node_name=alt_node_name)
 
         if hasattr(self.node_set, 'skip'):
             self._ast['skip'] = self.node_set.skip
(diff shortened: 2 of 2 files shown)


aanastasiou:
@eyalgr Hello and thank you for the commit. Can we please clarify how exactly does ``has``, works here? If it replicates the functionality of [``in``](https://neomodel.readthedocs.io/en/latest/queries.html#node-sets-and-filtering), we should keep only one of them. But if it is closer to [this](https://github.com/neo4j-contrib/neomodel/issues/284), that would be even better.

All the best

eyalgr:
@aanastasiou Hi, sure, please tell me if I'm wrong but both of these filter nodes based on the **properties of the node**, whereas in this commit `has` filters the nodes based on the **properties of the targets of the node's relations**, which as far as I know wasn't possible without two distinct queries in neomodel before.

Just to clarify, the implicit models in my examples look something like:
``` python
class Person(StructuredNode):
    pet =  RelationshipTo('Pet', 'PET')
    child =  RelationshipTo('Person', 'CHILD')
    name = StringProperty()

class Pet(StructuredNode):
    name = StringProperty()

```

aanastasiou:
Hello @eyalgr and thank you for the clarification.

Alright, this sounds like an additional useful feature. Can we please clarify the following?

1. This ``has``, seems to make it possible to query for a structure starting from a Node of some label. Can we please clarify how does this translate to cypher queries? Because at the moment, it looks like every subsequent ``has`` triggering a query itself (on every top level node)...and this can quickly get you into more trouble than writing down an equivalent parametrisable query in cypher and executing it with ``cypher_query`` (?)

2. >Node naming convention in QueryBuilder was modified - underscores in node names are duplicated as an escaping mechanism.

   This is kind of "scary" :) Can you please clarify if you foresee breaking any backwards compatibility?

All the best

eyalgr:
I’ll do my best to clarify, let me know if I misunderstood something -

1. In this implementation, only one cypher query will be issued to the db regardless of the amount of NodeSets involved - note you need to call .all or .get to trigger an actual query. In this implementation, the .has call instantiate a QueryBuilder for the provided NodeSet argument, pulls its ast and uses it to extend its own ast (https://github.com/neo4j-contrib/neomodel/pull/460/files#diff-eeaf1d19783ad16393cac96402adeeadR348)

2. It’s a good question. I think it can only break code that do nasty things with private members, such as the _ast, or tries to manually intervene with the resulting nodeset query. I should note that this commit does exactly that, but I feel like it’s ok because it’s limited to classes inside match.py. It may be a good idea to define a class-name-to-node-name function.

aanastasiou:
@eyalgr 

From a functionality point of view I think it's a useful operator to have.

There are some "caveats" that would have to be mentioned in the documentation I think. For example, to not actually call ``.all()`` or ``.get()`` on the intermediate ``NodeSets``.

Can you please write a bit about that in a brief documentation section about ``.has()``?

After that, it would be best to make sure that ``.has()`` is at least backwards compatible as a plain simple relationship check and that the "juggling" that ``has()`` is doing is...not going to break anything :)

All the best

eyalgr:
Note that passing the result of .all or .get to .has raises a ValueError.

I believe it is backwards compatible, the code flow for boolean arguments is left untouched and the tests are passing. Would you prefer to have tests for the new code too?

Where should I put the documentation?

aanastasiou:
@eyalgr Regarding the documentation: Any section you think it is suitable. I try to review changes in the documentation anyway.

eyalgr:
@aanastasiou 

While writing the documentation I noticed a bug in this feature, where a single node returns more than once in the query. I fixed that in a new change in this pull request:
https://github.com/neo4j-contrib/neomodel/pull/460/commits/bee75bd69d1addd6c5adfddc3eed72133e4585e3
This is very small but critical change, I would like to have your opinion on that.

Either way, the documentation is ready too.

The tests are failing in the initiation phase, not because of the code.

eyalgr:
Hi @aanastasiou, any news regarding this?

eyalcye:
Hi @aanastasiou, any updates?

Issue #301, opened 2018-02-21
Docs on exception raised by Structurednode.nodes.get is incorrect


mjmare:
Page http://neomodel.readthedocs.io/en/latest/getting_started.html mentions that:
```
# raises Person.DoesNotExist if no match
jim = Person.nodes.get(name='Jim')
```
So a class specific exception should be raised. Actually it does not. It raises: neomodel.core.DoesNotExist

```
institute = Institute.nodes.get(uuid=uuid)
  File "/Users/mjm/Dropbox/Devel/ACHN/Sociale_kaart_neomodel/venv3/lib/python3.6/site-packages/neomodel/match.py", line 511, in get
    raise self.source_class.DoesNotExist(repr(kwargs))
neomodel.core.DoesNotExist: {'uuid': '776dfd0d89dc44dca6x65160472ce6f0'}
```

aanastasiou:
@mjmare Thank you, this has been addressed and will appear corrected in the next release.

PR #465, opened 2019-08-09
Updating deprecated method `search()` to `filter()`

elena:
Responding to:
`DeprecationWarning: search() is now deprecated please use filter() and exclude()`


elena:
o, i c:
https://github.com/neo4j-contrib/neomodel/issues/301

aanastasiou:
@elena *"o, i c"* ... Is that "Oh, I see?" or something else I might not be getting here regarding 301?

whatSocks:
out of date per https://github.com/neo4j-contrib/neomodel/issues/301

PR #475, opened 2019-10-28
some cosmetic enhancements

jvalhondo:
Cosmetic enhancements following `flake8` linter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 sudo: required
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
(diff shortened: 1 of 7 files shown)


aanastasiou:
Thank you @jvalhondo . If you have some time to spare, please see [here](https://github.com/neo4j-contrib/neomodel/wiki/TODOs-&-Enhancements) for a set of known features to be added to `neomodel`.

jvalhondo:
Sure @aanastasiou , I've checked it out and I'm willing to get a deeper understanding of the library in order to contribute and add features.

It would be great to eventually catch up on a call, so you can give some guidance.

mariusconjeaud:
We are planning to add some pre-commit hooks, which would include a linter and a prettifier. Then there will be a first run, so this will become stale.

PR #479, opened 2019-11-12
Fix for docker tests

kupuguy:
The file `docker-compose.yml `is no longer generated dynamically, this means you can test on a single version (set in `.env`) simply with `docker-compose up`.

The Python/Neo4j versions may now be set in environment variables to run a subset of the
test environments.

e.g. For Neo4J 3.3 and 3.4 and Python 3.7 and 3.8 only:
```
export NEO4J_VERSIONS="3.4 3.3"
export PYTHON_VERSIONS="3.7 3.8"
./tests-with-docker-compose.sh
```

The docker image now uses the Python slim image instead of alpine as alpine has problems with `geos`.

I also fixed the test skipping so spatial tests actually run (the version comparison was taking 3.4 as less than 3.4.0).


aanastasiou:
@kupuguy Again, this sounds great, thank you very much.

whatSocks:
@kupuguy - could you rebase this branch and see if things still work as expected?

kupuguy:
> @kupuguy - could you rebase this branch and see if things still work as expected?

Rebased. I updated the Python and neo4j versions to just the supported versions.

MattGalvis:
I approve @aanastasiou

MattGalvis (approved):
looks good

ficast (approved):

PR #480, opened 2019-11-20
Add support for slow query log

StevenMapes:
This PR enhances the logging within the cypher_query method in ```neomodel.utils.py``` so that logging can be enabled only for queries that take longer than a specified amount of time. This is useful as the Neo4j query log is not available on the community edition and sometimes you only wish to log the slow queries and not everything especially if you are working on a production environment.

The logging uses a new optional environmental variable **NEOMODEL_SLOW_QUERIES** which can be set to a string whose value is thew threshold at which you want to start recording queries.

E.G If you set the value to the following ```os.environ["NEOMODEL_SLOW_QUERIES"] = "0.09"``` then only queries taking more than 0.09 seconds will be logged.

If the variable is not set, then it defaults to 0 and acts as before.

### DIFF
--- a/neomodel/util.py
+++ b/neomodel/util.py
@@ -243,8 +243,9 @@ class Database(local, NodeClassRegistry):
                                          retry_on_session_expire=False)
             raise
 
-        if os.environ.get('NEOMODEL_CYPHER_DEBUG', False):
-            logger.debug("query: " + query + "\nparams: " + repr(params) + "\ntook: {:.2g}s\n".format(end - start))
+        tte = (end - start)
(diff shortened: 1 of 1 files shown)


aanastasiou:
@StevenMapes Thank you very much that's great.

whatSocks:
@StevenMapes can you rebase your commits so we can test this with the latest updates on master?

StevenMapes:
> @StevenMapes can you rebase your commits so we can test this with the latest updates on master?

How's that?

MattGalvis (approved):

MattGalvis (approved):

ficast (approved):

Issue #488, opened 2020-02-05
CREATE UNIQUE no longer supported in Neo4j 4.0


JVemmer:
In relation to #487, I did a bit of testing with `cypher.default_language_version=3.5` set on the server. Simple queries seem to work, as long as they do try to connect any relationships. If they do, the following error comes up:

> File "/home/username/Projects/test/core/test.py", line 42, in try_relationship_connect
    n.things.connect(x)
File "/home/username/.local/lib/python3.8/site-packages/neomodel/relationship_manager.py", line 24, in checker
    return fn(self, *args, **kwargs)
File "/home/username/.local/lib/python3.8/site-packages/neomodel/relationship_manager.py", line 99, in connect
    self.source.cypher(q, params)
File "/home/username/.local/lib/python3.8/site-packages/neomodel/core.py", line 388, in cypher
    return db.cypher_query(query, params)
File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 33, in wrapper
    return func(self, *args, **kwargs)
File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 229, in cypher_query
    raise exc_info[1].with_traceback(exc_info[2])
File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 213, in cypher_query
    results, meta = [list(r.values()) for r in response], response.keys()
File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 213, in <listcomp>
    results, meta = [list(r.values()) for r in response], response.keys()
File "/home/username/.local/lib/python3.8/site-packages/neo4j/__init__.py", line 935, in records
    self._session.fetch()
File "/home/username/.local/lib/python3.8/site-packages/neo4j/__init__.py", line 516, in fetch
    detail_count, _ = self._connection.fetch()
File "/home/username/.local/lib/python3.8/site-packages/neobolt/direct.py", line 419, in fetch
    return self._fetch()
File "/home/username/.local/lib/python3.8/site-packages/neobolt/direct.py", line 461, in _fetch
    response.on_failure(summary_metadata or {})
File "/home/username/.local/lib/python3.8/site-packages/neobolt/direct.py", line 755, in on_failure
    raise CypherError.hydrate(**metadata)
neobolt.exceptions.CypherSyntaxError: CREATE UNIQUE is no longer supported. Please use MERGE instead (line 1, column 60 (offset: 59))
[...]

Issue #487, opened 2020-01-30
Replace {var} style variables with $var


KevinvOosterhout:
Since Neo4j 3.0 {var} style variables have been deprecated. This has been replaced by $vars. All neo4j 3.x versions prefer this style and neo4j 4.x versions require this style of variables. It's however not compatible with neo4j < 3.0.  

=== EDIT ===
This is required for 4.0 support

aanastasiou:
Thank you, will take a look at this soon

aanastasiou:
Had a first scan on tackling this and from this, the changes seem to be straightforward. 

The bulk of those changes are in modules `match.py`, `match_q.py`, `relationship.py` and `relationship_manager.py` and a couple in `util.py`.

aanastasiou:
@KevinvOosterhout Can you please post the complete trace that motivated this report?

JVemmer:
@aanastasiou I am not the original poster, but I created a test database to get a full traceback. (currently facing this issue as well)

> Test.nodes.get(uid='1234')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/username/.local/lib/python3.8/site-packages/neomodel/match.py", line 551, in get
    result = self._get(limit=2, lazy=lazy, **kwargs)
  File "/home/username/.local/lib/python3.8/site-packages/neomodel/match.py", line 542, in _get
    return self.query_cls(self).build_ast()._execute(lazy)
  File "/home/username/.local/lib/python3.8/site-packages/neomodel/match.py", line 447, in _execute
    results, _ = db.cypher_query(query, self._query_params, resolve_objects=True)
  File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 33, in wrapper
    return func(self, *args, **kwargs)
  File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 229, in cypher_query
    raise exc_info[1].with_traceback(exc_info[2])
  File "/home/username/.local/lib/python3.8/site-packages/neomodel/util.py", line 212, in cypher_query
    response = session.run(query, params)
  File "/home/username/.local/lib/python3.8/site-packages/neo4j/__init__.py", line 503, in run
    self._connection.fetch()
  File "/home/username/.local/lib/python3.8/site-packages/neobolt/direct.py", line 419, in fetch
    return self._fetch()
  File "/home/username/.local/lib/python3.8/site-packages/neobolt/direct.py", line 461, in _fetch
[...]

PR #490, opened 2020-02-19
Fix for #487

JVemmer:
As requested by @aanastasiou in #487, here is a fix for the param issue. Only explicit traversal seems to not work, but I have not focused on that, since it seems separate from the param issue (also possible that I simply do not understand it).
The following things were tested:
- [x] .get()
- [x] .save()
- [x] .filter()
- [x] .delete()
- [x] .refresh()
- [x] lazy evaluation (lazy=True)
- [x] property reference (n.id)
- [x] retrieving related nodes (n.friends_with.all())
- [ ] relationship connection (n.friends_with.connect(m)) - Does not work due to #488.
- [x] .exclude() (n.friends_with.exclude(name='test2'))
- [x] relationship disconnect (n.friends_with.disconnect(m) + n.friends_with.disconnect_all())

### DIFF
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -263,7 +263,7 @@ class StructuredNode(NodeBase):
             ", ".join("{0}: params.create.{0}".format(getattr(cls, p).db_property or p) for p in cls.__required_properties__))
         if relationship is None:
             # create "simple" unwind query
-            query = "UNWIND {{merge_params}} as params\n MERGE ({0})\n ".format(n_merge)
+            query = "UNWIND $merge_params as params\n MERGE ({0})\n ".format(n_merge)
         else:
             # validate relationship
             if not isinstance(relationship.source, StructuredNode):
@@ -275,8 +275,8 @@ class StructuredNode(NodeBase):
             from .match import _rel_helper
 
             query_params["source_id"] = relationship.source.id
-            query = "MATCH (source:{0}) WHERE ID(source) = {{source_id}}\n ".format(relationship.source.__label__)
-            query += "WITH source\n UNWIND {merge_params} as params \n "
+            query = "MATCH (source:{0}) WHERE ID(source) = $source_id\n ".format(relationship.source.__label__)
+            query += "WITH source\n UNWIND $merge_params as params \n "
             query += "MERGE "
             query += _rel_helper(lhs='source', rhs=n_merge, ident=None,
                                  relation_type=relation_type, direction=relationship.definition['direction'])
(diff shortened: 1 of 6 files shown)


aanastasiou:
@JVemmer 

Thanks for taking the time to work on these issues. Unfortunately, on my setup, I get a number of test failures due to errors with constraints.

I can see that you have requested me to review these changes. But there is a huge amount of checks that is performed by the tests themselves. Once you have made a change, please run the tests and the errors (or sometimes messages) originating from those tests will let you know what might be going wrong, *even though the code might appear to work as expected for certain functionality*.

This was the reason I asked if you could try running the tests a few weeks ago.

Now, I cannot run the tests on my setup (still looking into it), but looking at [the output of the tests that run automatically upon submitting a PR](https://travis-ci.org/neo4j-contrib/neomodel/jobs/652447970?utm_medium=notification&utm_source=github_status), I can see the following:

```
self = <neobolt.direct.Response object at 0x7fa82c6c3e10>
metadata = {'code': 'Neo.ClientError.Statement.ParameterMissing', 'message': 'Expected a parameter named nodeid'}
```

This occurs within `test_relationship_models.py`.

Can I please ask you to have another look at the substitutions you did and check if, perhaps mistakenly, you omitted a formatted variable?

To run the tests:

1. Usually, you create a virtual Python environment and perform your tests in that one.

2. Upon creating the environment, you could try installing neomodel to it directly. I usually take the following steps:
    1. Create the environment
    2. `pip install ipython pytest shapely` # Notice the `pytest` package right here.
    3. In the directory that neomodel's `setup.py` file is, run `pip install ./ -e`
        * This step will install `neomodel` in such a way that any changes you apply to its code will be immediately available to any code (anywhere within that environment) that uses `neomodel`.

At the end of this process, go to the same directory that neomodel's `setup.py` directory is and type `pytest`.

Many thanks again for taking the time to look into these issues anyway.

JVemmer:
@aanastasiou
I located the problem; was a bit too heavy handed in my replacements (mistook a format string for a params call). That is fixed now, and the tests all passed on my local machine. However, looking at the travis-ci tests being run here, there appeared a segfault at one of the tests? Unsure if that is code-related, or related to the travis setup itself.

aanastasiou:
@JVemmer I understand, thank you very much, will take a look at this soon.

JVemmer:
Hi @aanastasiou,
I kept getting confused about which branch I was dealing with, as well as which issue I was trying to solve, while trying fix the various issues I found. So I have made a decision to combine everything into a single pull request instead, so we both only have to look at a single thing, instead of the current 3 (which would have reached ~10 if everything was submitted individually). I will close this pull request to avoid confusion, and open up a single new one.

PR #499, opened 2020-03-12
Update for Neo4j 4.0

JVemmer:
Hi @aanastasiou,

This is the combined pull request I mentioned in #490, #491 and #489.
It contains a bunch of changes, focused primarily on getting neomodel to play nice with Neo4j >= 4.0.
It also fixed a few tests that were no longer functional, or were not working in the first place (in any version of Neo4j I could find).
This was tested with the [supported versions of Neo4j](https://neo4j.com/developer/kb/neo4j-supported-versions/) (at the time of writing that is 3.4, 3.5 and 4.0), both via the test framework and in production. All tests passed.
This pull request fixes #486, #487, #488 and #489.
It may also have a significant impact on #485.

**Security consideration:** Neo4j 4.0 no longer adds a self-signed certificate, and therefore one of the commits here disables ENCRYPTED_CONNECTION. This has no meaningful effect on local instances (running via localhost), but can have significant security impact when submitting things via network. I would recommend at least updating the documentation to highlight this, and in the long term investigate whether we can do anything to alleviate this.
**Notes for people using this in production systems:** You can re-enable ENCRYPTED_CONNECTION as soon as you have created your own certificate, and ensured that it is trusted by your machine. So this is not just a full drop of encrypted connections, merely Neo4j no longer doing the work for you itself (discussions on self-signed certificates and security are out of the scope here).

PR #503, opened 2020-05-12
Minor fix in the update for Neo4j 4.0

davidmoseler:
This solves a dependency issue in pull request #499

https://github.com/neo4j-contrib/neomodel/pull/499

The circular dependency issue was between core.py and util.py. I moved the import into the method that uses core inside util.py. I don't know if this is the best possible solution, but it makes it possible to run the code.

Obs: maybe I'm ignorant on github workflows, but I opened a new pull request since I don't know if it is possible to include a commit in someones else pull request without having repository write permissions.


JVemmer:
I cherry-picked your commit, so it's part of the #499 pull request :)

davidmoseler:
Thanks! I'll close the pull

Issue #441, opened 2019-04-24
Date properties are stored as string instead of temporal


machosec:
Neomodel's date property types (`DateProperty`, DateTimeProperty` and `DateTimeFormatProperty`) are inflated and stored as string values in neo4j instead of temporal values (https://neo4j.com/docs/cypher-manual/current/syntax/temporal/).
Using string instead of date/time types ruin many filtering/sorting capabilities which I guess are not desired.

aanastasiou:
@machosec I think that what you may be pointing to here is that `neomodel` does not use a standardised datetime format when storing dates and/or times (?). 

With server-side queries, you can always use the functions that cast a date bearing string to a date if you wish to extract specific parts of the date (e.g. in a "group by month" scenario) and at client-side, the date has already been turned into a `datetime` during filtering.

Is there a more specific example of any failures this might be leading to?

machosec:
Hi @aanastasiou ,
Neo4j version 3.4 introduced new temporal types and functions to better deal with dates. 
You can see the example below where the `date()` function (https://neo4j.com/docs/cypher-manual/current/functions/temporal/date/) returns a temporal type that is treated differently than just a string type. New temporal functions like `duration()` won't work on a date represented as a string.

![image](https://user-images.githubusercontent.com/17451964/56651338-7c1bd280-6691-11e9-8ade-b1ca4e720069.png)

aanastasiou:
@machosec I understand.

That would be the difference between `strptime()` and `neo4j.v1.types.temporal.Date`, etc at [this point](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/properties.py#L426) and of course, the inverse for `deflate()`. I would have to look more closely the data types that the `neo4j-driver` returns for dates as it might already be working with straightforward `datetime`s too.

Unfortunately, the `DateTimeFormatProperty` might not be as easily handled if it is impossible to represent partial dates.

Feel free to submit a PR if you can handle this.

machosec:
The `neotime` module used by `neo4j-driver` support the new temporal types as described here: https://neo4j.com/docs/api/python-driver/current/types/temporal.html
I can try to change the `inflate` and `deflate` logic of `DateProperty` and `DateTimeProperty` to properly store temporal types in neo4j without breaking the existing implementation

aanastasiou:
[...]

PR #512, opened 2020-08-12
support Neo4j 3.4, 3.5 explicitly

a-takahashi223:
neomodel supports DateTime and Spatial property so it seems to support Neo4j 3.4 and 3.5.
In fact, Travis CI includes Neo4j 3.4.
But the documentation says "supporting neo4j 3.0, 3.1, 3.2, 3.3."

On the other hand, neomodel tests failed with Neo4j 3.5 as seen in #454 , #415 , #393 .
But I figured out that it was caused by sharing connection between subprocesses in the multiprocess test.
I fixed it in this PR. Now neomodel tests passes with Neo4j 3.5.

So I think we can say clearly neomodel supports Neo4j 3.4 and 3.5.

Note: the test failure with Neo4j 3.5 was caused by sharing connection so you can fix it by changing `multiprocessing` start method to `"spawn"` or `"forkserver."`
But `"spawn"` and `"forkserver"` do not work well in Travis CI, I don't know why. It looks like the bug of Travis CI.


elferia:
Sorry, as mentioned in #441 , neomodel can not handle Neo4j's temporal property introduced in 3.4.
But still we can say that neomodel support Neo4j 3.4, 3.5 in spite of such limitations.

a-takahashi223:
@aanastasiou @robinedwards Could you give me your opinion about this PR?

whatSocks:
Will merge this branch then move on to addressing the issues in the issue list

PR #530, opened 2020-11-22
support native neo4j datetime type

yeus:
Made this little patch in order to support native neo4j datetime format https://neo4j.com/docs/api/python-driver/current/temporal_types.html#neo4j.time.DateTime.

here is the documentation about this format. I used some functions that don't appear in the official documentation. especially:

`neo4j.to_native()`

I also don't know the neomodel policy about where to import the neo4j library.


whatSocks:
Hello - thank you for your contribution. We are working on updating this repo and reviewing PRs but, to set expectations, maintenance is moving at a slow pace.

ficast (changes requested):
@yeus can you add a test for this? Take a look on `test/test_properties.py ` for examples. Sorry about the delay, and thanks for that!

- [ ] add unit test

yeus:
> @yeus can you add a test for this? Take a look on `test/test_properties.py ` for examples. Sorry about the delay, and thanks for that!
> 
>     * [ ]  add unit test

will do!

aanastasiou:
Hello @yeus , do you think that this could also potentially address #610 too?

yeus:
> Hello @yeus , do you think that this could also potentially address #610 too?

It kind of looks like that could be the case. to be sure I would test it though...

mariusconjeaud (changes requested):
Add some tests

mariusconjeaud:
Replaced by #811

Issue #526, opened 2020-11-04
4.0 support for multi database


praveenguduri:
Neo4j python driver has added support for multi database. Is there any parameter in neomodel to set the database for session or driver connection?

adnanmuttaleb:
Anything new for this issue?

PR #556, opened 2021-04-30
Add support for multi databases

mRoca:
Resolve https://github.com/neo4j-contrib/neomodel/issues/526

Driver database documentation: https://neo4j.com/docs/api/python-driver/4.1/api.html#database

With this PR, we can specify the database name in the neo4j or bolr URL:

`bolt://neo4j:neo4j@localhost:7687` => default database
`bolt://neo4j:neo4j@localhost:7687/mydb` => `mydb` database

### DIFF
--- a/doc/source/configuration.rst
+++ b/doc/source/configuration.rst
@@ -6,7 +6,7 @@ This section is covering the Neomodel 'config' module and its variables.
 Database
 --------
 
-Seting the connection URL::
+Setting the connection URL::
 
     config.DATABASE_URL = 'bolt://neo4j:neo4j@localhost:7687`
 
@@ -18,6 +18,10 @@ Adjust connection pool size::
 
     config.MAX_CONNECTION_POOL_SIZE = 50  # default
 
+Setting the database name, for neo4j >= 4::
+
+    config.DATABASE_URL = 'bolt://neo4j:neo4j@localhost:7687/mydb`
+
 Enable automatic index and constraint creation
 ----------------------------------------------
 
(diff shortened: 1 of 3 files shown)


mRoca:
It would be easier to avoid the new ENV variable, and allowing to pass the DB name into the url: `bolt://neo4j:neo4j@localhost:7687/mydb`, for instance.

=> What do you think about that ?

nossila:
It looks good to me.

> It would be easier to avoid the new ENV variable, and allowing to pass the DB name into the url: bolt://neo4j:neo4j@localhost:7687/mydb, for instance.

I like the idea a lot, but seems like this is something to be done on neo4j-driver instead. But I'd say lets use your solution now

mRoca:
> It looks good to me.
> 
> > It would be easier to avoid the new ENV variable, and allowing to pass the DB name into the url: bolt://neo4j:neo4j@localhost:7687/mydb, for instance.
> 
> I like the idea a lot, but seems like this is something to be done on neo4j-driver instead. But I'd say lets use your solution now

If we only use the `config.DATABASE_URL = 'bolt://neo4j:neo4j@localhost:7687/mydb'` syntax, the solution would be more flexible, as we won't have to update all related packages (as `django-neomodel` for instance): updating the connection string will be enough to use this features, without having to create another PR with the new `DATABASE_NAME` config var to handle.

nossila:
@mRoca actually I think it makes a lot of sense to have the database in the URL like you suggested, I was under the impression that the URL parsing was done by the `neo4j-driver` but turns out it is not :P https://github.com/neo4j-contrib/neomodel/blob/bfeae3e10498006dc0aae6bab2885a3c0f194dc3/neomodel/util.py#L90

mRoca:
I've reworked this PR. Now, the database is extracted from the url: `bolt://neo4j:neo4j@localhost:7687/mydb` if provided.

`bolt://neo4j:neo4j@localhost:7687` still works, with the default database.

mRoca:
@nossila What do you think about this new version?

Who can review this PR?

PR #559, opened 2021-05-17
Create python-publish.yml

whatSocks:


ficast (approved):

nossila (approved):

python-publish.yml
```diff
@@ -0,0 +1,31 @@
+# This workflow will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
```
alimcmaster1, line 8: You will want to do this when the release is actually published right?!

alimcmaster1 (changes requested):

python-publish.yml
whatSocks: we're still stuck on administering this portion so this PR is going to stay open for an undetermined time. 

        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}

alimcmaster1: We should add them as secrets to the repo. This look good for uploading wheels/sdist.  We can use the API_TOKEN if easier & upload to test pypi to check all is working here. https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-the-distribution-to-pypi-and-testpypi

Shout if you need help happy to contribute.

mariusconjeaud:
I fixed this and the pipeline ran fine for version 4.0.10

Closing this PR since I started from the GitHub's latest sample.

PR #585, opened 2021-11-04
Fix DateProperty inflate

ralfeus:
`DateProperty` requires `datetime` data type and in `deflate()` it's converted to isoformat of view `YYYY-MM-DDTHH:MM:SS`. Then `inflate()` tries to convert it back to `datetime` by using `strptime(.., "%Y-%m-%d")`, which doesn't deal with 'THH:MM:SS'. Python since 3.7 has `fromisoformat()`, which is a counterpart for `isoformat()`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ development.env
 .ropeproject
 \#*\#
 .eggs
+bin
+lib
+.vscode
+pyvenv.cfg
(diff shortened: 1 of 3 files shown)


olegchigirin:
In this case, neomodel can't support python 3.6- anymore. Currently, it requires python 3.5+. It will affect users with old versions of python.

PR #586, opened 2021-11-15
Fix DateProperty inflate

ralfeus:
DateProperty requires datetime data type and in deflate() it's converted to isoformat of view YYYY-MM-DDTHH:MM:SS. Then inflate() tries to convert it back to datetime by using strptime(.., "%Y-%m-%d"), which doesn't deal with 'THH:MM:SS'. 

Fixed with removing time part of the string. Tested on Python 3.6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ development.env
 .ropeproject
 \#*\#
 .eggs
+bin
+lib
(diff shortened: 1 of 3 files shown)


whatSocks:
thank you

PR #587, opened 2021-11-16
Use modern python API for metadata

alimcmaster1:
Ref: https://docs.python.org/3.9/library/importlib.metadata.html#module-importlib.metadata

Ref: https://github.com/googleapis/python-api-core/issues/27#issuecomment-676563754
(@pganssle is CPython Core Dev. :))

pkg_resources is causing a few issues for us, particularly in lambda envs

This is a much better way in Python 3.8+

cc @whatSocks @nossila

### DIFF
--- a/neomodel/__init__.py
+++ b/neomodel/__init__.py
@@ -22,4 +22,12 @@ __author__ = 'Robin Edwards'
 __email__ = 'robin.ge@gmail.com'
 __license__ = 'MIT'
 __package__ = 'neomodel'
-__version__ = pkg_resources.get_distribution('neomodel').version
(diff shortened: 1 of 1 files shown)


neomodel/__init__.py
```diff
@@ -22,4 +22,12 @@
 __email__ = 'robin.ge@gmail.com'
 __license__ = 'MIT'
 __package__ = 'neomodel'
-__version__ = pkg_resources.get_distribution('neomodel').version
+if (sys.version_info.major, sys.version_info.minor) >= (3, 8)
```
alimcmaster1, line 25: This will work fine until a new major python version - which is fine for now

Issue #392, opened 2018-12-04
relationship inheritance


DachuanZhao:
Here is the code:
```
class FriendRel(StructuredRel):
    since = DateTimeProperty(
        default=lambda: datetime.now(pytz.utc)
    )
    met = StringProperty()

class BestFriendRel(FriendRel):
    pass

class Person(StructuredNode):
    name = StringProperty()
    friends = RelationshipTo('Person', 'FRIEND', model=FriendRel)
    best_friends = RelationshipTo('Person', 'BESTFRIEND', model=BestFriendRel)

rel = jim.best_friends.connect(bob)
rel.since # datetime object
```
May I use relationship inheritance like this?

DachuanZhao:
Thanks a lot~
It's difficult for me to contribute that......

Issue #130, opened 2014-10-05
allShortestPaths


maxtepkeev:
Hi!

Thanks for the great package!

I'm trying to get allShortestPaths between two nodes with the following cypher query:

``` python
results, _ = db.cypher_query(
    "MATCH "
    "(from:Point {system_name: 'Point1'}), "
    "(to:Point {system_name: 'Point10'}), "
    "p = allShortestPaths(from-[:NEXT|CONNECTS_WITH*..100]-to) RETURN p"
)
print([Point.inflate(row[0]) for row in results])
```

During `db.cypher_query` execution the following exception is raised:

```
Traceback (most recent call last):
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/neomodel/util.py", line 195, in cypher_query
    results = self._execute_query(query, params)
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/neomodel/util.py", line 189, in _execute_query
    result = neo4j.CypherResults(cq._cypher._post({'query': query, 'params': params or {}}))
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/py2neo/neo4j.py", line 1139, in __init__
    for row in content["data"]
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/py2neo/neo4j.py", line 1139, in <listcomp>
    for row in content["data"]
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/neomodel/util.py", line 77, in _hydrated
    return type(data)([_hydrated(datum) for datum in data])
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/neomodel/util.py", line 77, in <listcomp>
    return type(data)([_hydrated(datum) for datum in data])
  File "/usr/local/projects/neo4j/python/lib/python3.4/site-packages/neomodel/util.py", line 75, in _hydrated
    raise NotImplemented("Don't know how to inflate: " + repr(data))
TypeError: 'NotImplementedType' object is not callable
```

Am I doing something wrong ?

aanastasiou:
@maxtepkeev, @richardschwab 

This is an interesting feature for a number of reasons but while it is impossible to do exactly what you are after at the moment, writing a response for it made me realise a short and long term improvement we could have in neomodel.

So, the situation is this: Indeed, at the moment we are not resolving the `Path` data type which is what the Neo4j driver would return from the database for the sort of queries you indicate there. So:

[...]

PR #592, opened 2021-11-24
Adding object resolution for Relationship type objects

nossila:
This PR is based on https://github.com/neo4j-contrib/neomodel/pull/421 with the intention of resolving conflicts and get it merged.

> This would make it possible to address #130 and #392.


olegchigirin:
The problem with the new approach is when people build their logic on the assumption that the same relationship class could use any relationship label.

In our case, we can't even start our project after this commit. We have do refactor at least half of relationship definitions in our  code to be ready for this update 

It's dangerous to add such logic as a minor library version. Also it doesn't mentioned in changelog

PR #604, opened 2022-02-04
Fix #600 unexpected RelationshipClassRedefined for siblings with same label

### LINKED ISSUE #600: Pull request #592 cause an error in case of relationship inharitance
Approved pull request #592 is not a minor change. For it makes our application completely impossible to start.

In <4.0.8 release, such code was valid:

```
class RelA(StructuredRel):
    some_property = 'some value'

class ModelA(StructuredNode):
     rel_a: neomodel.ZeroOrMore = RelationshipTo('ModelA', 'REL_TYPE_A', model=RelA)
     rel_b: neomodel.ZeroOrMore = RelationshipTo('ModelA', 'REL_TYPE_B', model=RelA)                                                           
```

It is useful when describing different relationships with the same relationship logic.

After the 4.0.8 release, such relationship definitions raise exceptions:

`Relationship of type REL_TYPE_B redefined as <class 'RelA'>`

This is not a minor code update, and in our case, it requires refactoring at least half of the relationship definitions.

**Update:**

I investigate what is going on. The problem above is not a correct explanation of the error I got in my project. The real issue is the following:

Let us have relationship classes with the following inheritance

```
class Class1(StructuredRel):
    pass


class SubClass1(Class1):
    pass


class SubClass2(Class1):
    pass
```

And following model class:

```
class RelationshipDefiner(StructuredNode):
    rel_1 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=Class1)
    rel_2 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=SubClass1)
    rel_3 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=SubClass2)
```

We get the following error:

```
neomodel.exceptions.RelationshipClassRedefined: Relationship of type SOME_REL_LABEL redefined as <class 'SubClass1'>.
RelationshipDefiner --> <class 'RelationshipDefiner'>
SOME_REL_LABEL --> <class 'SubClass1'>
```

The correct behavior is to allow both of them to use 'SOME_REL_LABEL' because they extend the same base class.

Error is coming from here:

https://github.com/neo4j-contrib/neomodel/blob/2850013f01f5c51ef19fc43323fed8aecb97090c/neomodel/relationship_manager.py#L378

```
if not issubclass(model, db._NODE_CLASS_REGISTRY[label_set]):
    raise RelationshipClassRedefined(relation_type, db._NODE_CLASS_REGISTRY, model)
else:
    db._NODE_CLASS_REGISTRY[label_set] = model
```
Because we replace the previous mentioning in case the next one is the subclass of the first one.

I see the correct approach of this: 

We have to store in _NODE_CLASS_REGISTRY the most parent class from all extended types, described this relationship. In my case, we have to keep Class1.

**Update 2**:

```
class RelationshipDefiner(StructuredNode):
    rel_2 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=SubClass1)
    rel_3 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=SubClass2)
    rel_1 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=Class1)
```

An additional case is this one. When the parent class is not the first one, and the first two classes using 'SOME_REL_LABEL' are on the same level in the hierarchy, we got a RelationshipClassRedefined error on trying to process rel_3.

nossila:
Fix issue #600

### DIFF
--- a/neomodel/relationship_manager.py
+++ b/neomodel/relationship_manager.py
@@ -24,6 +24,12 @@ def check_source(fn):
         return fn(self, *args, **kwargs)
     return checker
 
+# checks if obj is a direct subclass, 1 level
+def is_direct_subclass(obj, classinfo):
+    for base in obj.__bases__:
+        if base == classinfo:
+            return True
+    return False
 
 class RelationshipManager(object):
     """
@@ -375,8 +381,11 @@ class RelationshipDefinition(object):
                 # If the relationship mapping exists then it is attempted to be redefined so that it applies to the same
                 # label. In this case, it has to be ensured that the class that is overriding the relationship is a
                 # descendant of the already existing class
-                if not issubclass(model, db._NODE_CLASS_REGISTRY[label_set]):
-                    raise RelationshipClassRedefined(relation_type, db._NODE_CLASS_REGISTRY, model)
+                model_from_registry = db._NODE_CLASS_REGISTRY[label_set]
+                if not issubclass(model, model_from_registry):
+                    is_parent = issubclass(model_from_registry, model)
+                    if is_direct_subclass(model, StructuredRel) and not is_parent:
+                        raise RelationshipClassRedefined(relation_type, db._NODE_CLASS_REGISTRY, model)
                 else:
                     db._NODE_CLASS_REGISTRY[label_set] = model
             except KeyError:
--- a/test/test_issue600.py
+++ b/test/test_issue600.py
@@ -1,0 +1,70 @@
+"""
+Provides a test case for issue 600 - "Pull request #592 cause an error in case of relationship inharitance".
+
+The issue is outlined here: https://github.com/neo4j-contrib/neomodel/issues/600
+"""
+
+import os
+import neomodel
+import datetime
+import pytest
+import random
+
+try:
+    basestring
+except NameError:
+    basestring = str
+
+class Class1(neomodel.StructuredRel):
+    pass
+
+class SubClass1(Class1):
+    pass
+
+class SubClass2(Class1):
+    pass
+
+class RelationshipDefinerSecondSibling(neomodel.StructuredNode):
+    rel_1 = neomodel.Relationship('RelationshipDefinerSecondSibling', 'SOME_REL_LABEL', model=Class1)
+    rel_2 = neomodel.Relationship('RelationshipDefinerSecondSibling', 'SOME_REL_LABEL', model=SubClass1)
+    rel_3 = neomodel.Relationship('RelationshipDefinerSecondSibling', 'SOME_REL_LABEL', model=SubClass2)
+
+
+class RelationshipDefinerParentLast(neomodel.StructuredNode):
+    rel_2 = neomodel.Relationship('RelationshipDefinerParentLast', 'SOME_REL_LABEL', model=SubClass1)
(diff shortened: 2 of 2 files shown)


nossila:
@olegchigirin please take a look at this solution, I'm not 100% sure if this will cover all use cases so if you think of any other tests I can add there to make sure the solution covers other possible cases I would appreciate it a lot.

nossila:
@olegchigirin you described in issue #600:

```python
class RelationshipDefiner(StructuredNode):
    rel_1 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=Class1)
    rel_2 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=SubClass1)
    rel_3 = Relationship('RelationshipDefiner', 'SOME_REL_LABEL', model=SubClass2)
```

> The correct behavior is to allow both of them to use 'SOME_REL_LABEL' because they extend the same base class.

But in your last comment you said that the following definition should raise and error: 

```python
class RelationshipDefinerSecondSibling(neomodel.StructuredNode):
    rel_2 = neomodel.Relationship('RelationshipDefinerSecondSibling', 'SOME_REL_LABEL_2', model=SubClass1)
    rel_3 = neomodel.Relationship('RelationshipDefinerSecondSibling', 'SOME_REL_LABEL_2', model=SubClass2)
    rel_1 = neomodel.Relationship('RelationshipDefinerSecondSibling', 'SOME_REL_LABEL_1', model=Class1)
```

Is it because Class1 (which is the parent class for both SubClass1 and 2) is using a different label?
What if you never define a relationship to Class1 in you node? 

Right now both those cases passes the tests, but I'm having issues understanding the case where it should error. I think to achieve that we would need to re-validate the definitions after all definitions are done to be able to validate and throw an error when using the same label to 2 different classes/models that extends same base class.

olegchigirin:
In the second case, we have a bit different problem.

Children's classes could use relationship labels different from a parent class. So, if a child declares its relationship label, only it and its children could use it.

Different between the first and second cases is the following: Children declare the same relationship to parents in the first case. And it is allowed; in the second case, children declare the same relationship different from parent one, and it is forbidden.

@whatSocks please correct me if I incorrectly understand the relationship rules you implemented

nossila:
@olegchigirin to be honest I'm having a hard time understanding why the second case would be forbidden, don't seems like neo4j itself has this constrain, it actually does not allow multiple labels in the same relationship so it is not able to know anything about our python's classes inheritances as far as I know.

> in the second case, children declare the same relationship different from parent one, and it is forbidden.

I think we should allow children to have different labels from its parent, this seems like the correct way to allow different relationship models, different labels are different relationships and the model (StructuredRel) just [describes the properties of that relationship](https://neomodel.readthedocs.io/en/latest/relationships.html#properties).

Can you elaborate a little more please?

whatSocks:
@olegchigirin  perhaps asking a different way - is the solution in this PR "good enough for now"? It seems your requirements are more strict than what is present in Neo4j, and others might have more permissive requirements.

olegchigirin:
@whatSocks seems like something that has to work. Later we can improve it.

Sorry for the long response, I had to evacuate.

mariusconjeaud:
Included in release 4.0.9

PR #607, opened 2022-02-21
Upgrade Shapely dependency to 1.8.1

eddyfabery:
----
[Shapely Releases ](https://github.com/shapely/shapely/releases)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
         'pytest-runner'] if any(x in ('pytest', 'test') for x in sys.argv) else [],
     tests_require=['pytest>=6.2.4', 'shapely', 'neobolt'],
     install_requires=['neo4j-driver==4.3.6', 'pytz>=2021.1',
-                      "neobolt==1.7.17", "Shapely==1.7.1", "six==1.16.0"],
+                      "neobolt==1.7.17", "Shapely==1.8.1", "six==1.16.0"],
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         'Intended Audience :: Developers',


whatSocks:
@olegchigirin @eddyfabery 

being paranoid but you're saying that this update will allow users to use python 3.10?

olegchigirin:
Yeah, sorry.

Update of shapely allows to use python 3.10

nossila:
@eddyfabery could you also drop support for Python 3.5 on `setup.py` and `.travis.yml`? I think it makes sense since Shapely 1.8.0 drops support for python 3.5 as well.

joekohlsdorf:
Is it necessary to hardcode the version? This could cause conflicts.

setup.py
```diff
@@ -20,7 +20,7 @@
         'pytest-runner'] if any(x in ('pytest', 'test') for x in sys.argv) else [],
     tests_require=['pytest>=6.2.4', 'shapely', 'neobolt'],
     install_requires=['neo4j-driver==4.3.6', 'pytz>=2021.1',
-                      "neobolt==1.7.17", "Shapely==1.7.1", "six==1.16.0"],
+                      "neobolt==1.7.17", "Shapely==1.8.1", "six==1.16.0"],
```
joekohlsdorf, line 23: Is it necessary to pin the version at all? This could cause conflicts.
```suggestion
                      "neobolt==1.7.17", "Shapely>=1.7.1,<1.9", "six==1.16.0"],
```

joekohlsdorf (changes requested):

joekohlsdorf:
Note that tests_require allows any version of shapely so why is it pinned at all?

setup.py
louisphilippe2: > Is it necessary to pin the version at all? This could cause conflicts.

I pray this will be accepted soon so i can use neomodel with python 3.10

alimcmaster1:
> Note that tests_require allows any version of Shapely and this code section shows that it is optional, why is it pinned/force installed at all? https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/contrib/spatial_properties.py#L26

Agree with this! It seems like Shapely is really an optional dependency of neomodel, since alot of functionality can be used without it. Similar to optional dependencies in pandas. https://pandas.pydata.org/docs/getting_started/install.html

I would suggest we put "Shapely>=1.7.1,<1.9" in an extras_require section.

alimcmaster1:
> @eddyfabery could you also drop support for Python 3.5 on `setup.py` and `.travis.yml`? I think it makes sense since Shapely 1.8.0 drops support for python 3.5 as well.

Agree think this should be a separate PR I can take a look.

alimcmaster1:
Fixed by: https://github.com/neo4j-contrib/neomodel/pull/612/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R22

mariusconjeaud:
Included in release 4.0.9

PR #610, opened 2022-03-02
Update match.py

15m43lk4155y:
Filtering a node set on a date property doesn't work, I traced it back to the process_filter_args function in match.py where the DateProperty deflate method returns iso format which results in the matching query parameter value being a string instead of a datetime.date object therefore it doesn't match it, I handled it with a condition on the DateProperty but it can be handled by changing the deflate method's return as well

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -1,10 +1,9 @@
 from .core import StructuredNode, db
-from .properties import AliasProperty, DateProperty
+from .properties import AliasProperty
 from .exceptions import MultipleNodesReturned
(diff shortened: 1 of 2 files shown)


alimcmaster1:
Thanks for the contribution please could you write a test case to help demonstrate the issue?

alimcmaster1 (changes requested):
Ref above

mariusconjeaud:
Closing for lack of activity and requested change not adressed, plus potentially fixed by #530. Feel free to re-open it with more information / content.

PR #612, opened 2022-03-23
Update Shapely

alimcmaster1:
Ref: https://github.com/neo4j-contrib/neomodel/pull/607

cc. @[joekohlsdorf](https://github.com/joekohlsdorf)

From the code looks it seems like Shapely is really an optional dependency of neomodel, since alot of functionality can be used without it. Similar to optional dependencies in pandas. https://pandas.pydata.org/docs/getting_started/install.html
Ref: https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/contrib/spatial_properties.py#L26

@whatSocks @nossila to review

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,9 +18,10 @@ setup(
              'scripts/neomodel_remove_labels'],
     setup_requires=[
         'pytest-runner'] if any(x in ('pytest', 'test') for x in sys.argv) else [],
-    tests_require=['pytest>=6.2.4', 'shapely', 'neobolt'],
+    tests_require=['pytest>=6.2.4'],
+    extras_require={'extras': ['Shapely>=1.7.1,<1.9']},
     install_requires=['neo4j-driver==4.3.6', 'pytz>=2021.1',
-                      "neobolt==1.7.17", "Shapely==1.7.1", "six==1.16.0"],
+                      "neobolt==1.7.17", "six==1.16.0"],
     classifiers=[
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -18,9 +18,10 @@
              'scripts/neomodel_remove_labels'],
     setup_requires=[
         'pytest-runner'] if any(x in ('pytest', 'test') for x in sys.argv) else [],
-    tests_require=['pytest>=6.2.4', 'shapely', 'neobolt'],
```
joekohlsdorf, line 21: Is it ok to remove `neobolt` from the test requirements?

setup.py
```diff
@@ -18,9 +18,10 @@
...
+    tests_require=['pytest>=6.2.4'],
```
joekohlsdorf, line 21: Is it ok to remove `neobolt` from the test requirements?

alimcmaster1: neobolt is an install requirement - so will be installed from there :) test requirements should generally be additional requirements just needed for testing e.g pytest etc..

alimcmaster1:
Friendly ping to maintainers here - @whatSocks @nossila to review

alimcmaster1:
From speaking to maintainers it would seem this change is correct. Shapely is really an optional dependency and not a hard requirement as per the comment: https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/contrib/spatial_properties.py#L26

PR #616, opened 2022-04-07
improve RelationshipDefinition constructor performance

mudetz:
Provides performance improvements for changes introduced in #603
Instead of loading the whole stack frame at once, these changes allow incrementally loading frames as needed.
In general, only the 3-6 outermost frames are needed.

We tested the changes from #603 in a large project and saw the time that takes to instantiate `RelationshipDefinition` go up for up to 10 seconds from a baseline of milliseconds. This observations were made at import time, where the stack trace is huge.

### DIFF
--- a/neomodel/relationship_manager.py
+++ b/neomodel/relationship_manager.py
@@ -363,10 +363,18 @@ class RelationshipManager(object):
 
 class RelationshipDefinition(object):
     def __init__(self, relation_type, cls_name, direction, manager=RelationshipManager, model=None):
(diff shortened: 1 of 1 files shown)


alimcmaster1:
Thanks for the contribution :) Is this code tested somewhere in unit test in this project? Do we have code coverage here?

alimcmaster1 (changes requested):
Ref above

joekohlsdorf:
This is covered by existing relationship tests, it is just a change to make already existing code more efficient.
We have been running this patch in production since opening the PR with no issues.

Issue #633, opened 2022-08-13
Achieve parity of `StructuredNode.cypher` and `db.cypher_query`


AntonLydike:
I noticed that `StructuredNode.cypher` is missing the `resolve_objects` argument. I looked around and found no reason why it should be missing. While adding support for it, I noticed that there are quite a few more flags on `db.cypher_query` that are not available for `StructuredNode.cypher`. Are there any reasons for that?

I am specifically talking about:

 - `handle_unique` Whether or not to raise UniqueProperty exception on Cypher's ConstraintValidation errors
 - `retry_on_session_expire` Whether or not to attempt the same query again if the transaction has expired

aanastasiou:
Hello @AntonLydike `db.cypher_query()` is what every other component uses to actually communicate with the DBMS.

A trivial example is using that function to return "raw" results as opossed to resolved objects (e.g, imagine a query that returns raw strings and integers that cannot be resolved to any of the classes in a given hierarchy). The various parameters of `db.cypher_query()` are there to serve different purposes depending on where that function is called from.

`StructuredNode.cypher()` also ultimately calls `db.cypher_query()` but in turn, it is called from operations that don't really need to return classes (or at least, did not, at the time it was created). For example, when updating a `StructuredNode`, all we need are the node's properties (which are returned in the driver's object) to update a specific `StructuredNode` attributes.

Was there a particular use case you had in mind when proposing these changes?

mariusconjeaud:
Closing for cleaning the Issues log and Thanos' answer looks OK to me. Feel free to re-open if you had a specific use case in mind requiring changes.

PR #634, opened 2022-08-13
added resolve_objects argument to StructuredNode.cypher

AntonLydike:
This is the first step as outlined in #633 to reach feature parity between StructuredNode.cypher and db.cypher_query.

I added a test, although I am not quite sure if I did everything right... The query in the test returns a weird list-in-list structure:
```pyhton3
data, meta = jim.cypher("MATCH (a) WHERE id(a)=$self OR id(a)=$cat RETURN a", {'cat': cat.id}, resolve_objects=True)
print(data)
# prints something like [[<User2 ...>], [<Animal ...>]]
```

I flattened the list, but that might not be intended behavior.

### DIFF
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -401,7 +401,7 @@ class StructuredNode(NodeBase):
         results = db.cypher_query(query, params)
         return [cls.inflate(r[0]) for r in results[0]]
 
-    def cypher(self, query, params=None):
+    def cypher(self, query, params=None, resolve_objects=False):
         """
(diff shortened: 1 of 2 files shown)


AntonLydike:
I would like to add documentation, but that would clash with my previous PR about documentation #631, if that's merged first I can rebase this branch and add documentation.

mariusconjeaud:
@AntonLydike PR #631 has been merged if you want to take this up again

Issue #533, opened 2020-12-02
Enable multiple relation labels `(a)-[r:TYPE1|TYPE2]->(b)` in builder


AntonLydike:
As per [documentation](https://neomodel.readthedocs.io/en/latest/relationships.html#explicit-traversal) you can supply the explicit relationship traversal settings with `relation_type` of `None|*|<name>`:

> ```python
> definition = dict(node_class=Person, direction=OUTGOING,
>                  relation_type=None, model=None)
>relations_traversal = Traversal(jim, Person.__label__,
>                                definition)
>all_jims_relations = relations_traversal.all()
>```
>The defintion argument is a mapping with these items:
>node_class | The class of the traversal target node.
>-- | --
>direction | match.OUTGOING / match.INCOMING / match.EITHER
>relation_type | Can be None (for any direction), * for all paths or an explicit name of a relation type (the edge’s label).
>model | The class of the relation model, None for such without one.

My suggestion is to extends this to include a list of names. This would result in a query using the `(a)-[:TYPE1|TYPE2]->(b)` syntax described by the neo4j docs [here](https://neo4j.com/docs/cypher-manual/3.5/syntax/patterns/#cypher-pattern-relationship). This neo4j feature is supported since at least version 3.5, so compatibility should be of no concern.

This would be an example usage:

```python
definition = dict(node_class=Person, direction=INCOMING,
                  relation_type=('FATHER','MOTHER'), model=None)
relations_traversal = Traversal(jim, Person.__label__,
                                definition)
all_jims_children = relations_traversal.all()
```

I am currently working on a PR, feedback is appreciated.

aanastasiou:
@AntonLydike I have not seen PRs yet but yes this would be great and I think fairly straightforward. The challenging bit might be instantiating and returning the right relationship type.

AntonLydike:
I have linked PR #534, it seems like the build system is broken at the moment (as described in the PR). Instantiating relationship types is not the problem as the query only returns the connected nodes, not edges.

It seems like this repo is a little bit dead tho. No activity on my PR and Issue in more than a month. I am a little disapointed tbh.

PR #635, opened 2022-08-13
WIP: Add multi-hop support to RelationshipDefinition

AntonLydike:
I want to add support for multiple-hop relationship definitions. For example:

```
Class Person(StructuredNode):
  name = StringProperty()
  
  parent = RelationshipTo('Person', 'PARENT')
  ancenstors = RelationshipTo('Person', 'PARENT', n_hops='*')
  decendants = RelationshipFrom('Person', 'PARENT', n_hops='*')
```

I am open for ideas and pointers!

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -14,7 +14,7 @@ except NameError:
     basestring = str
 
 
-def _rel_helper(lhs, rhs, ident=None, relation_type=None, direction=None, relation_properties=None, **kwargs):
+def _rel_helper(lhs, rhs, ident=None, relation_type=None, direction=None, relation_properties=None, n_hops=None, **kwargs):
     """
     Generate a relationship matching string, with specified parameters.
     Examples:
@@ -56,7 +56,12 @@ def _rel_helper(lhs, rhs, ident=None, relation_type=None, direction=None, relati
         stmt = stmt.format('[*]')
     else:
         # explicit relation_type
-        stmt = stmt.format('[{0}:`{1}`{2}]'.format(ident if ident else '', relation_type, rel_props))
+        stmt = stmt.format('[{0}:`{1}`{2}{3}]'.format(
+            ident if ident else '',
+            relation_type,
+            n_hops if n_hops else '',
+            rel_props,
+        ))
 
     return "({0}){1}({2})".format(lhs, stmt, rhs)
 
--- a/neomodel/relationship_manager.py
+++ b/neomodel/relationship_manager.py
@@ -362,7 +362,7 @@ class RelationshipManager(object):
 
 
 class RelationshipDefinition(object):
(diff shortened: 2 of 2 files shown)


AntonLydike:
Question: How to best handle calls to `connect()`, `disconnect()`, `reconnect()`, etc?

I have added a property called `allows_creation` which marks if this relationship manager is allowed to create edges. The code then checks for this whenever necessary, and if it isn't allowed to create an edge, it throws a NotImplementedError (this type is probably improvable!)

The intended behaviour for delete and delete_all is now to delete the whole path. I am unsure if it is desirable to have such a foot gun though.

AntonLydike:
The `allows_creation` property could then be used to make re-implementing #533 way easier as this ran into similar issues back then.

aanastasiou:
@AntonLydike 

A few thoughts in no particular order:

1. You propose a Relationship type that can include more than one edges.

2. In a typical Neo4j database, the underlying property graph is assumed to be a [Multigraph](https://en.wikipedia.org/wiki/Multigraph). The only thing that Multigraphs add is the ability to have more than one edge between any two nodes.

3. A Relationship that *spans more than one nodes* is akin to a hyper-edge in a [hypergraph](https://en.wikipedia.org/wiki/Hypergraph).

4. A neomodel Relationship does not have hyper-edge semantics.

5. While **it is possible** to model hyper-edges in Neo4j, I would be hesitant in trying to lump hyper-edges with Relationships.
    * For example, Neo4j includes the concept of [a path](https://neo4j.com/docs/cypher-manual/current/introduction/uniqueness/) but that is not a primitive data type throughout CYPHER. Node connections (from the point of view of a data type) are still assumed to be edges with properties that link *always* two nodes.
    * Furthermore, the example you provide here assumes that all segments of a hyper-edge (all the "hops" of a "Relationship") are supposed to have the same property model (the same Relationship class definition). This is not always the case. A hyper-edge with properties can have different attributes at its segments. It is an entirely different data type.

Therefore, I would like to propose the following:

* Neomodel could include a ``contrib`` entity to model hyper-edges. This entity would have to be a new **type**. This new type could possibly inherit from Relationship but also include all the functionality that is required of hyper-edges.

* To decide what functionality should be assigned to hyper-edges, you should have one or more motivating use-cases. 
   * Here for example, your motivating use case seems to be being able to address a set of nodes with specific characteristics given a starting node. In my mind, this still sounds like it would be better served via a query. But you may have come across a use-case that indicates otherwise. We need to see the use-cases and the examples before setting off to incorporate this in neomodel.

My perception is that this feature needs more thinking before we can consider a PR, but I am happy to think twice :)

mariusconjeaud:
I think this idea should be rethought in the context of the 5.4.x new `traverse_relations` and `fetch_relations` methods ; and also with the recent Neo4j QPP feature

PR #642, opened 2022-09-17
Fixes #277 relationship match return node set

### LINKED ISSUE #277: document error: match() returns Traversal
The `match()` method used to filter on relationship properties returns `NodeSet` according to [the document](http://neomodel.readthedocs.io/en/latest/module_documentation.html#neomodel.relationship_manager.RelationshipManager.match).
But actually it returns `Traversal` (so you cannot call the methods of `NodeSet`, for example, `get()`.)

AntonLydike:
This is an attempt at fixing #277. I am not sure if such a breaking change would be justified though. Looking at the code, I definitely get the feeling, that it's *supposed* to return a NodeSet.

This probably has to wait for the next minor or even major release for that reason...

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -783,14 +783,14 @@ class Traversal(BaseSet):
         self.name = name
         self.filters = []
 
-    def match(self, **kwargs):
+    def match(self, **kwargs) -> NodeSet:
         """
         Traverse relationships with properties matching the given parameters.
 
             e.g: `.match(price__lt=10)`
 
         :param kwargs: see `NodeSet.filter()` for syntax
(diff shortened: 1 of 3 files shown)


aanastasiou:
@AntonLydike This looks like a good candidate to incorporate in the next round of changes. Is there any reason not to?

mariusconjeaud:
OK, this has been a looooong time coming, sorry about that.

I finally checked, this is still relavant ; but cannot be rebased sanely as the code has drifted since then. So I will have to close this PR, but I am writing a different PR to include it in release 7.0.0. Thanks @AntonLydike

PR #644, opened 2022-09-26
fix remove_all_labels() failure with Neo4j 4.3+

### LINKED ISSUE #567: [Support Neo4j 4.3] Neo4j 4.3: list index out of range in drop_indexes
After upgrading to neo4j 4.3 I'm experiencing the following error when executing remove_all_labels()

```
File "/usr/local/lib/python3.9/dist-packages/neomodel/core.py", line 81, in remove_all_labels
  drop_indexes(quiet=False, stdout=stdout)
File "/usr/local/lib/python3.9/dist-packages/neomodel/core.py", line 62, in drop_indexes
  index[7][0], index[8][0]))
IndexError: list index out of range
```

Is neomodel intended to be compatible with neo4j 4.3? From the tests it seems that it is only tested up to to neo4j 4.1

a-takahashi223:
fix #567


neomodel/core.py
```diff
@@ -57,6 +57,11 @@ def drop_indexes(quiet=True, stdout=None):
             stdout.write(' - Dropping index on label {0} with property {1}.\n'.format(
                 match.group(1), match.group(2)))
         else:
+            # Token lookup indexes (introduced in Neo4j 4.3)
+            # They are created automatically so they should not be dropped
+            if not index[7]:
+                assert not index[8], "Token lookup indexes have neither label nor property"
```
olegchigirin, line 63: maybe ValueError is better? 

```suggestion
                if not index[8]:
                    continue
                else:
                     raise ValueError(f"Token {index[8]} Token lookup indexes have neither label nor property")
```

setup.py
```diff
@@ -18,7 +18,7 @@
              'scripts/neomodel_remove_labels'],
     setup_requires=[
         'pytest-runner'] if any(x in ('pytest', 'test') for x in sys.argv) else [],
-    tests_require=['pytest>=6.2.4'],
+    tests_require=['pytest>=6.2.4', 'pytest>=6.2.4,<7.1.0;python_version<"3.7"'],
```
olegchigirin, line 21: A bit confused about this. Why do we need a version less than "3.7"?

neomodel/core.py
olegchigirin, line 63: assert is dangerous in case you run a project with optimization:

https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement
https://docs.python.org/3/using/cmdline.html#cmdoption-O

If someone runs the project with the -O flag you will never get this exception and always reproduce the bug.

Please, replace it for the value error

olegchigirin (changes requested):

neomodel/core.py
AntonLydike: Doesn't `not index[7]` still throw the same exception, as we are accessing the array out of bounds? Shouldn't it be `len(index) < 7` instead?

AntonLydike:
Hi @a-takahashi223, thanks for your contribution! I'll take a look at this as soon as I find some time!

setup.py
a-takahashi223: pytest>=7.1.0 dropped support for python 3.6. So if you run `python setup.py test` with python<=3.6, the constraint `"pytest<7.1.0` is necessary.

neomodel/core.py
a-takahashi223: As shown in https://github.com/neo4j-contrib/neomodel/issues/567#issuecomment-877568650, `index[7]` and `index[8]`, which are lists of the labels and the names of the properties, respectively, are empty for token lookup indexes.
So `index[7]` and `index[8]` are valid, `index[7][0]` and `index[8][0]` are invalid.

a-takahashi223: replaced with `ValueError` 4c2e46a

Issue #651, opened 2022-11-22
Documentation: Notes for StringProperty link to Google


kl-thamm:
On this page (https://neomodel.readthedocs.io/en/latest/properties.html): There is a link with notes for the StringProperty, which just link to Google.

mariusconjeaud:
I merged the PR ; but not sure yet how the documentation is published.

PR #656, opened 2023-01-19
Fix link in documentation

martin-mueller-cemas:
This PR fixes a link on the property types page. Currently the text leads to `http://www.google.com` while it probably should lead to the Notes section where the notes are.


a-takahashi223:
related: #651

PR #661, opened 2023-02-01
Upgrade to support Neo4j 4.4 and Python 3.10

mariusconjeaud:
- Update neo4j driver to 4.4.10
- Confirmed support of all 4.x Neo4j versions (up to LTS version 4.4)
- Update shapely to 1.8.1 to support Python 3.9 and 3.10
- Drop support of Python 3.6
- Up pytest requirement to 7.1 (Python 3.7+)


olegchigirin:
>Update shapely to 1.8.1 to support Python 3.9 and 3.10

It seems like it was fixed in the current main branch because we upgraded our project based on neomodel to python 3.11, which works well. 

Are some game-breaking with upgrading the codebase to support neo4j 5+?

PR #667, opened 2023-02-15
Fixes  #666 install labels Neo4j v5.4

### LINKED ISSUE #666: Invalid constraint syntax, ON and ASSERT should not be used.
## Expected Behavior (Mandatory)
when you run the command to install labels. It was supposed to create them.

## Actual Behavior (Mandatory)
But it will return this error. 
The error message you  suggests that the syntax you used to create the constraint is no longer valid in the version of Neo4j 5. The ON and ASSERT keywords were used in older versions of Neo4j to create constraints, but they have since been replaced by FOR and REQUIRE.

## How to Reproduce the Problem
just use Neo4j 5 and current neomodel

### Simple Example
I was using django_neomodel, when I create the models.py. I was supposed to run the install_labels command. So, when I ran the command I got the error.

## Screenshots (where it's possibile)
Screenshot from 2023-02-14 21-04-02

Currently used versions
neo4j 5.4.0
django_neomodel 0.0.7
current version of neomodel

### Versions

  - OS:
  - Library:
  - Neo4j:

hamashoool:
when creating labels, I replace the old `ON` and `ASSERT` keywords were used in older versions of Neo4j to create constraints, but they have since been replaced by `FOR `and `REQUIRE`.

### DIFF
--- a/neomodel/core.py
+++ b/neomodel/core.py
@@ -115,7 +115,7 @@ def install_labels(cls, quiet=True, stdout=None):
                 stdout.write(' + Creating index {0} on label {1} for class {2}.{3}\n'.format(
                     name, cls.__label__, cls.__module__, cls.__name__))
             try:
-                db.cypher_query("CREATE INDEX on :{0}({1}); ".format(
+                db.cypher_query("CREATE INDEX for :{0}({1}); ".format(
                     cls.__label__, db_property))
             except ClientError as e:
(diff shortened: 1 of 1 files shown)


hamashoool:
also it will be part of #664.

mariusconjeaud:
See PR #668 . It was ready but I hadn't opened it yet. It includes your changes but also a few more things, including updating the tests

PR #668, opened 2023-02-15
Upgrade/neo4j 5 #664

mariusconjeaud:


mariusconjeaud:
Note : Some tests fail after merging the recent changes to the build into this branch. This looks like it might be due to the install/remove label scripts not being run.

neomodel/core.py
```diff
@@ -22,18 +22,27 @@ def drop_constraints(quiet=True, stdout=None):
...
+
+    results_as_dict = [dict(zip(meta, row)) for row in results]
+    for constraint in results_as_dict:
+        # Versions prior to 4.3 returned a 'description' column
+        if "description" in constraint:
+            db.cypher_query('DROP {!s}'.format(constraint["description"]))
+
+            pattern = re.compile(':(.*) \).*\.(\w*)')
+            match = pattern.search(constraint["description"])
+            stdout.write(''' - Dropping unique constraint and index on label {0} with property {1}.\n'''.format(
```
aanastasiou: Would it be worth starting to drop the ` "".format() ` way of doing string interpolation in favour of f-strings?

I would suggest that we go ahead with f-strings from now on and make it a task to turn every string interpolation to an f-string for a future release.

test/test_label_drop.py
```diff
@@ -13,36 +13,73 @@ class ConstraintAndIndex(StructuredNode):
...
         else:
-            db.cypher_query('CREATE ' + constraint[1])
-    for index in indexes_before:
+            constraint_type_clause = "UNIQUE"
+            if constraint["type"] == "NODE_PROPERTY_EXISTENCE":
+                constraint_type_clause = "NOT NULL"
+            elif constraint["type"] == "NODE_KEY":
+                constraint_type_clause = "NODE KEY"
+            try:
+                db.cypher_query('CREATE CONSTRAINT {0} FOR (n:{1}) REQUIRE n.{2} IS {3}'
```
aanastasiou: This would definitely read better as an f-string (Nothing to be done now, just a note for future reference)

test/test_models.py
```diff
@@ -94,6 +94,7 @@ def test_save_to_model():
 
 
 def test_unique():
+    install_labels(User)
     User(email='jim1@test.com', age=3).save()
     try:
         User(email='jim1@test.com', age=3).save()
```
aanastasiou, line 100: This is a round-about way of checking that an exception was raised. Pytest has a more readable way of doing this:

```
with pytest.raises(neomodel.exceptions.ConstraintValidationFailed) as e:   
    User(email='jim1@test.cim', age=3).save()
```

**EDIT:** Maybe the exception is `UniqueProperty` here but you get the idea.

neomodel/core.py
```diff
@@ -162,6 +162,7 @@ def __new__(mcs, name, bases, namespace):
         namespace['DoesNotExist'] = \
             type(name + 'DoesNotExist', (DoesNotExist,), {})
         cls = super(NodeMeta, mcs).__new__(mcs, name, bases, namespace)
+        cls.DoesNotExist._model_class = cls
```
aanastasiou, line 189: Sorry, out of my own curiosity, why has this not caused a problem before?

neomodel/core.py
```diff
@@ -117,7 +117,7 @@ def install_labels(cls, quiet=True, stdout=None):
                 stdout.write(' + Creating unique constraint for {0} on label {1} for class {2}.{3}\n'.format(
                     name, cls.__label__, cls.__module__, cls.__name__))
             try:
-                db.cypher_query("CREATE CONSTRAINT "
+                db.cypher_query("CREATE CONSTRAINT constraint_unique_{0}_{1} "
```
aanastasiou, line 121: Very useful

aanastasiou (approved):
This looks great. That odd `try / catch` might be one of those things that we queue for a future upgrade along with turning all string interpolations to f-strings.

neomodel/core.py
olegchigirin: Until neomodel supports python 3.7, it is not the case. But sad =(

aanastasiou: @olegchigirin I do not understand this comment, but there is nothing stopping us from making this change in an upcoming major release.

olegchigirin: Sorry, I was misleading.

f-string added in python 3.6, not python 3.8.

Yes, in this case f-string are useful staff to add

aanastasiou: @olegchigirin No worries, thank you for the clarification.

mariusconjeaud: Yes I agree, f-strings are better. And we can already do it because the current version of neomodel supports Python 3.7 -> 3.10

mariusconjeaud: Updated in upcoming commit

test/test_models.py
mariusconjeaud: Updated

PR #672, opened 2023-02-17
Corrected the misconfiguration of pyproject.toml

mariusconjeaud:
Location of package was set too deep in the project structure, resulting in package not being found.


mariusconjeaud:
Fixed integration tests

PR #675, opened 2023-02-23
Modified pyproject.toml to properly install the neomodel scripts

aanastasiou:
This PR solves two problems with `pyproject.toml` on `4.0.10`:

1. The version is brought back to "static" (to resolve problems with installing a local (dev) version); and more importantly
2. The `neomodel` scripts getting their proper names
    * In `4.0.10` they are installed as `install-labels`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     "neobolt==1.7.17",
     "six==1.16.0",
 ]
-dynamic = ["version"]
+version='4.0.10'
 
 [project.urls]
 documentation = "https://neomodel.readthedocs.io/en/latest/"
@@ -56,5 +56,5 @@ dev = [
 addopts = "--resetdb"
 
 [project.scripts]
-install-labels = "scripts.neomodel_install_labels:main"
-remove-labels = "scripts.neomodel_remove_labels:main"
+neomodel_install_labels = "scripts.neomodel_install_labels:main"
+neomodel_remove_labels = "scripts.neomodel_remove_labels:main"


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "neobolt==1.7.17",
     "six==1.16.0",
 ]
-dynamic = ["version"]
+version='4.0.10'
```
mariusconjeaud, line 41: This will change (and potentially break) the Publish Package pipeline behavior. It's OK, just need to be aware of it, or find a way around it.

pyproject.toml
```diff
@@ -56,5 +56,5 @@ dev = [
 addopts = "--resetdb"
 
 [project.scripts]
-install-labels = "scripts.neomodel_install_labels:main"
-remove-labels = "scripts.neomodel_remove_labels:main"
+neomodel_install_labels = "scripts.neomodel_install_labels:main"
```
mariusconjeaud, line 59: Does it still work with the latest renaming of the script files to "neomodel_install_labels.py" ? (See latest commit in master branch)

mariusconjeaud (changes requested):

pyproject.toml
aanastasiou: @mariusconjeaud the scripts have always been `neomodel_install_labels` and `neomodel_remove_labels` and any user scripts still using those names for management would fail.

This change just brings back the original names (and is tested and works)

mariusconjeaud: OK great ! Then we're all good.

mariusconjeaud (approved):

PR #677, opened 2023-02-27
Make slicing returns a NodeSet.

tonioo:
According to documentation:
https://neomodel.readthedocs.io/en/latest/queries.html?highlight=count#iteration-slicing-and-more

Slicing a NodeSet is supposed to return a NodeSet so it allows us to continue chaining calls.

The current implementation is actually returning a list. This fix is an attempt to make code consistent with documentation.

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -553,7 +553,7 @@ class BaseSet(object):
             elif key.start:
                 self.skip = key.start
 
-            return self.query_cls(self).build_ast()._execute()
+            return self
 
         elif isinstance(key, int):
             self.skip = key


aanastasiou:
@tonioo great catch, thank you for your contribution.

tonioo:
@aanastasiou You're welcome, thank you for maintaining neomodel.
What are the criteria to respect before such PR can be merged into master branch?

aanastasiou:
@tonioo All PRs from all contributors no matter how small or large are reviewed for inclusion in upcoming releases. We have a few changes queued for the next relase which might be within the next month (maybe sooner).

tonioo:
@aanastasiou Ok, thank you for the answer.

PR #686, opened 2023-03-22
New way to fetch additional relations in query.

tonioo:


mariusconjeaud:
Here is the first merge back from the NovoNordisk StudyBuilder fork.

There is some doc and tests that you can read to understand what it is about. In a few words, it allows to MATCH paths starting from your main (start) node, either assertive or optional.
This allows to (for instance) filter down nodes based on a path they should be on, or get additional properties located on other nodes along the path.

**Important note** : this PR stops at returning the resolved objects as they are returned in Cypher.
So for instance, if your path looks like (startNode:Person)-[r1]->(middleNode:Country)<-[r2]-(endNode:Continent), then you will get a list of results, where each result is a list of (startNode, r1, middleNode, r2, endNode). These will be resolved, so startNode will be a Person class as defined in neomodel for example.
We have developed a method to parse this as a tree of projected objects but it is outside the scope of this PR since the point is to project to Pydantic classes (to return from an API).
We have also discussed a way of projecting this Django-style using an object cache ; but this needs further discussion.

So, for now, I would suggest adding disclaimers in the docs that this is an initial stab at fetching relations along with your main node, but you have to do the final mapping of results yourself in the way that best suits you.
Happy to discuss this.

mariusconjeaud (approved):

PR #695, opened 2023-04-12
Remove deprecated methods

mariusconjeaud:
Deprecated methods since version 3 with deprecation notice :

- StructuredRel.delete() - _Replaced by .disconnect()_
- RelationshipManager.search() - _Replaced by .filter() and .exclude()_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ lib
 pyvenv.cfg
 coverage_report/
 .coverage*
+.DS_STORE
+cov.xml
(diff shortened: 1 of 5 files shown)


mariusconjeaud:
These two methods were the only ones marked @deprecated, but we need to search for other potential deprecated leftovers.

aanastasiou (approved):
This was indeed needed.

In a follow up commit, I have also removed the `_id` property and the importing of the `deprecated` decorator wherever it was imported without being used.

PR #713, opened 2023-07-07
Support for Node's optional labels

mariusconjeaud:
This allows to add an `__optional_labels__` array to define extra labels for a Node.

For example :

`class MyNodeLabel(StructuredNode):
    __optional_labels__=["ExtraLabel"]
`

Then this allows to create nodes with both :MyNodeLabel and :ExtraLabel


mariusconjeaud:
Test error not introduced by this branch ; it is a problem with Aura, which also now happens on other branches - investigating.

tonioo (approved):

Issue #715, opened 2023-07-10
Automatic inflation of Path nodes returned by cypher query


73VW:
## Guidelines

Please note that GitHub issues are only meant for bug reports/feature requests. 
Before sending unsolicited pull requests, please open a feature request to discuss your proposal.

If you have questions on usage, please ask on [the Discussion Forum](https://community.neo4j.com) instead of creating an issue here.

## Feature description (Mandatory)

Hello everyone, 

It seems that nodes contained in a path are not inflated when using `resolve_objects=True` in `db.cypher_query`.

Checking in the code seems to confirm that. [This method](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/util.py#L208) seems to inflate `Node`, `Relationship` or iterate over a `list` but doesn't do anything regarding `Path` objects returned by a cypher query.

My concern is that I make a query to find all nodes at a depth X from a certain center node. 

``` py
        ....
        query = f"""
    MATCH (p) WHERE id(p) = {center}
    MATCH path = (p)-[r:links*1..{depth}]-(connected)
    RETURN path
"""
        results, meta = db.cypher_query(query, resolve_objects=True)
        ...
```

This returns a list of `Path` objects.

## Considered alternatives

Iteratively inflate all nodes of a path.

## How this feature can improve the project?

This will resolve nodes of a path under the hood.

aanastasiou:
Indeed. These features have been developed and merged at different time scales, hence these discrepancies. Thanks for flagging this up.

mariusconjeaud:
Part of release 5.1.1

Issue #709, opened 2023-06-10
Feature Request: Multiple related node match in a single query


rmtobin:
## Feature description
It would be amazing if I was able to fetch a Node/NodeSet plus given related Nodes in a single query, ala Django's select_related/prefetch_related. Being able to do this would alleviate painful N+1 query problems that are very easy to run into using Neomodel's query system.

For example, say I want to populate a list of books with their authors and genres. I have `Book` nodes, off of which I have relationships to `Author` nodes, and `Genre` nodes. As it is, without manually writing a Cypher query, I have to first query all the `Book` nodes, then, for each `Book` node, fetch each related `Author` and `Genre` node separately. This guarantees now that I have to run 2N+1 queries, where N is the number of `Book` nodes (which in my case could be many hundreds, and I'm actually fetching more relationships than two).

Ideally, it'd be nice if Neomodel did something similar to Django's select_related for querysets, so instead I could do something like `Book.nodes.all().select_related('authors', 'genres')` and have Neomodel intelligently generate a single query to fetch everything all at once. It would be extra nice if it also had a NodeSet cache (like select_related) where it caches those results, so then subsequent queries like `book_node.genres.all()` within the same execution frame, where `book_node` was in the `NodeSet` returned in my earlier query, would be able to be fulfilled by the cache without another query to the database. This would make using Neomodel in the context of say a GraphQL server much less painful, among other benefits.

## Considered alternatives
Generally I've had to resort to hand-writing Cypher and managing my own rudimentary query cache, which defeats a lot of the point of using an ORM.

## How this feature can improve the project?
It would allow users to create more complex, more performant queries, and make it more tenable to use Neomodel for highly-related/high-cardinality, large volume data. I do realize this is a difficult and complex feature, but it would bring it closer to on par to many of the major SQL-based ORMs, which makes it that much easier to justify adopting Neo4j.

aanastasiou:
@rmtobin Hello and thanks for this proposal. 

Let me get this straight:

At the moment, what you would do is:

1. Retrieve the books with something like: `all_books = Book.nodes.all()`
2. Given a book, access some property like: `book_authors = all_books[0].authors.all()`

Or maybe (continuing from above):

```
book_data = [ (a_book, a_book.authors.all(), a_book.genres.all()) for a_book in all_books]
```

Which would "buffer" all related information about a book in separate entries in `book_data`.

[...]

Issue #473, opened 2019-10-07
Add typing annotations?


laurentS:
Hi all,
I've started using `neomodel` (and `neo4j`) a few weeks ago, and quickly realised that typing annotations would be a tremendous help to pick up a bunch of errors/typos/etc...

As I couldn't find any stubs, I got started writing my own version, but it's more work than I initially thoughts.
So I wanted to check:
- whether anyone else has already written annotations?
- whether anyone else would be interested in them?
- whether anyone with a better knowledge of the code base than mine would be willing to help a bit with some parts where I feel completely stuck.

Thanks for the great framework, by the way!

aanastasiou:
@laurentS Sounds like a great idea but possibly better suited for a major release. Happy to help.

laurentS:
I looked into this a little, using stubs so that I don't need to touch the code in `neomodel`.
For me, the most valuable part is typing for the properties on nodes/relations as a first goal. I'm only using python 3.6+, so my thoughts below may or may not work for older versions of python.
@aanastasiou it would be great to have your thoughts on the options below. It's a bit hackish, and I'm sure there's a better way to make this work!

With inspiration from [this SO answer](https://stackoverflow.com/a/54948342/1138710), I can think of a few options, none perfect so far:

## option 1: do not annotate properties in stubs, only in code
```python
# nothing in stubs

# in my code
class MyNode(StructuredNode):
    foo: str = StringProperty()  # type: ignore

m = MyNode()
m.foo = "some string"
reveal_type(m.foo)  # str
```
This works, and requires no changes to the `neomodel` codebase, but it means some of the magic is lost as the user of `neomodel` has to annotate all properties, plus silence mypy.

## option 2: annotate properties as "Any" in stubs, and in code
```python
# in stub
class StringProperty(Any): ...

# in my code
class MyNode(StructuredNode):
    foo: str = StringProperty()

m = MyNode()
m.foo = "some string"
reveal_type(m.foo)  # str
```
Similar to the above, we still need to annotate each property, but at least mypy doesn't complain on each line.

## Option 3: Fool mypy with extra annotations
```python
# in stub
class StringProperty(NormalizedProperty):
    ...
    def __get__(*args: Any) -> str: ...

# in my code
class MyNode(StructuredNode):
    foo = StringProperty()

m = MyNode()
m.foo = "some string"  # OK
m.foo = 42  # Incompatible types in assignment (expression has type "int", variable has type "str")
[...]

PR #716, opened 2023-07-10
Added automatic path inflation

73VW:
Implements #715 

I am not really convinced about the implementation but I couldn't manage to do better without rewriting half of the package.
I welcome discussion and comments on this pull request.

I have not implemented tests but they will come as soon as I have time.

I've run all the tests except those for "Aura" because I don't know what it is. They finished succesfully.

### DIFF
--- a/neomodel/util.py
+++ b/neomodel/util.py
@@ -12,6 +12,7 @@ from neo4j.api import Bookmarks
 from neo4j.debug import watch
 from neo4j.exceptions import ClientError, ServiceUnavailable, SessionExpired
 from neo4j.graph import Node, Relationship
+from neo4j.graph import Path as NeoPath
 
 from neomodel import config, core
 from neomodel.exceptions import (
@@ -61,6 +62,18 @@ def clear_neo4j_database(db, clear_constraints=False, clear_indexes=False):
         core.drop_indexes()
 
 
+class Path(NeoPath):
+    def __init__(self, nodes, *relationships):
+        for i, relationship in enumerate(relationships, start=1):
+            start = relationship.start_node()
+            end = relationship.end_node()
+            if start not in nodes and end not in nodes:
+                raise ValueError(
+                    "Relationship %d does not connect to all of the nodes" % i
+                )
+        self._nodes = tuple(nodes)
+        self._relationships = relationships
+
 class Database(local):
     """
     A singleton object via which all operations from neomodel to the Neo4j backend are handled with.
@@ -100,7 +113,10 @@ class Database(local):
             "neo4j+ssc",
         ]
 
-        if parsed_url.netloc.find("@") > -1 and parsed_url.scheme in valid_schemas:
+        if (
+            parsed_url.netloc.find("@") > -1
+            and parsed_url.scheme in valid_schemas
+        ):
             credentials, hostname = parsed_url.netloc.rsplit("@", 1)
             username, password = credentials.split(":")
             password = unquote(password)
@@ -132,7 +148,9 @@ class Database(local):
         self.url = url
         self._pid = os.getpid()
         self._active_transaction = None
-        self._database_name = DEFAULT_DATABASE if database_name == "" else database_name
+        self._database_name = (
+            DEFAULT_DATABASE if database_name == "" else database_name
+        )
 
         # Getting the information about the database version requires a connection to the database
         self._database_version = None
@@ -251,7 +269,58 @@ class Database(local):
             # The database server is not running yet
             pass
 
-    def _object_resolution(self, result_list):
+    def _object_resolution(self, object_to_resolve):
+        """
+        Performs in place automatic object resolution on a result
+        returned by cypher_query.
+
+        The function operates recursively in order to be able to resolve Nodes
+        within nested list structures and Path objects. Not meant to be called 
+        directly, used primarily by _result_resolution.
+
+        :param object_to_resolve: A result as returned by cypher_query.
+        :type Any:
+
+        :return: An instantiated object.
+        """
+        # Below is the original comment that came with the code extracted in
+        # this method. It is not very clear but I decided to keep it just in
+        # case 
+        #
+        #
+        # For some reason, while the type of `a_result_attribute[1]`
+        # as reported by the neo4j driver is `Node` for Node-type data
+        # retrieved from the database.
+        # When the retrieved data are Relationship-Type,
+        # the returned type is `abc.[REL_LABEL]` which is however
+        # a descendant of Relationship.
+        # Consequently, the type checking was changed for both
+        # Node, Relationship objects
+        if isinstance(object_to_resolve, Node):
+            return self._NODE_CLASS_REGISTRY[
+                frozenset(object_to_resolve.labels)
+            ].inflate(object_to_resolve)
+
+        if isinstance(object_to_resolve, Relationship):
+            return self._NODE_CLASS_REGISTRY[
+                frozenset([object_to_resolve.type])
+            ].inflate(object_to_resolve)
+
+        if isinstance(object_to_resolve, NeoPath):
+            new_nodes = []
+            for node in object_to_resolve.nodes:
+                new_nodes.append(self._object_resolution(node))
+            new_relationships = []
+            for relationship in object_to_resolve.relationships:
+                new_rel = self._object_resolution(relationship)
+                new_relationships.append(new_rel)
+            return Path(new_nodes, *new_relationships)
+
+        if isinstance(object_to_resolve, list):
+            return self._result_resolution([object_to_resolve])
+        return object_to_resolve
+
+    def _result_resolution(self, result_list):
         """
         Performs in place automatic object resolution on a set of results
         returned by cypher_query.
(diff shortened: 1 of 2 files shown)


neomodel/util.py
```diff
@@ -205,7 +230,41 @@ def rollback(self):
...
+            ].inflate(object_to_resolve)
+
+        if isinstance(object_to_resolve, NeoPath):
+            new_nodes = []
+            for node in object_to_resolve.nodes:
+                new_nodes.append(self._object_resolution(node))
+            new_relationships = []
+            for relationship in object_to_resolve.relationships:
+                new_rel = self._object_resolution(relationship)
+                print(new_rel)
```
aanastasiou, line 259: Please review the use of this `print` right here, probably a debugging remnant.

test/test_issue283.py
```diff
@@ -81,16 +81,22 @@ class SomePerson(BaseOtherPerson):
 
 
 # Test cases
-def test_automatic_object_resolution():
+def test_automatic_result_resolution():
```
aanastasiou, line 84: Why is this test getting renamed here?

neomodel/util.py
```diff
@@ -75,6 +80,21 @@ def __str__(self):
         return "\n".join(ncr_items)
 
 
+class Path(NeoPath):
+    def __init__(self, nodes, *relationships):
+        assert all([isinstance(node, BaseNode) for node in nodes])
+        for i, relationship in enumerate(relationships, start=1):
+            assert isinstance(relationship, BaseRel)
```
aanastasiou, line 87: This check, along with the one on line 85 are possibly redundant. What else than nodes and relationships can the retrieved objects be here?

(Also, this kind of check will at some point be superseded by [type checking](https://github.com/neo4j-contrib/neomodel/issues/473)

aanastasiou (changes requested):
Hello

That was very quick. Thank you for this PR. I think it is a good start, please see comments.

All the best

73VW:
Hey @aanastasiou!

Thanks for your review. Indeed that was a quick and dirty implementation because I needed it for one of my projects.

I will take a moment to add a few tests and fix the points you've suggested.

test/test_issue283.py
73VW: I have renamed both test because I have renamed the original method.

Instead of `_object_resolution`, I have renamed it to `_result_resolution` and in fact used `_object_resolution` for the part that.. resolves objects.

Check https://github.com/neo4j-contrib/neomodel/pull/716/files#diff-5a14848686f658a9fab17bad447b0d3631c13d078bd3bbde13804138e3150e24R267 and https://github.com/neo4j-contrib/neomodel/pull/716/files#diff-5a14848686f658a9fab17bad447b0d3631c13d078bd3bbde13804138e3150e24R233

Therefore, I have renamed the test methods to match with the name of the method it is covering. I can revert that if you want :)

neomodel/util.py
73VW: I drew a lot of inspiration from the implementation of the neo4j `Path` class: https://github.com/neo4j/neo4j-python-driver/blob/5.0/src/neo4j/graph/__init__.py#L297. This might not be the best idea, I don't know.
On the other hand, I couldn't test that they were `StructuredNode` and `StructuredRel` because that was causing a cyclic import error. That's the reason I made base classes for both types. 
I can remove them if you prefer.

73VW: Hey @aanastasiou,

Have you seen my reply?

Cheers

neomodel/util.py
mariusconjeaud: @73VW Can you remove the print or resolve the conversation, whichever is right ?

neomodel/util.py
mariusconjeaud: @aanastasiou This is the last standing point for this PR, any opinion on this ?

mariusconjeaud:
I resolved the merge conflicts for you. Waiting for the last two open points, then we can merge it.

73VW:
Hey @mariusconjeaud,

I've removed the useless `print` statement. I will check the rest of the the changes you've made when we get an answer from @aanastasiou. Is that ok for you?

aanastasiou:
@73VW ...Was there a question I have missed?

neomodel/util.py
73VW: I agree with removing this pseudo-type.

However I have no clue about the different db versions. I am completely new to neo4j and may need help identifying which versions should be tested. 

I haven't implemented any tests for the moment.

What should I do now?

aanastasiou: The tests are running once you push your PR. The whole battery of tests runs on each different version of neo4j.

Tests use `pytest`. Have a look at that. Also, the test needs a docker instance running the neo4j db of choice.

So, whether locally or when pushing the PR, the tests will run and we will be able to see the result.

In terms of what we are doing next:

1. Please go ahead and change the code as discussed
2. Is this more or less what you are trying to do?
    ```
          z = neomodel.db.cypher_query(" match p = (a:Blah)-[b:Foo]->(c:Goo) return p limit 1", resolve_objects= True)
    ```
This would return a Path object on `z`.

If that is the general idea then I can write the tests, if there are more details that I have missed please write here a minimal example of how to make use of the specific code you have contributed.

Do you think we can do this?

73VW: Hey @aanastasiou,

I have made the changes according to your request. 

@mariusconjeaud 
I have rebuilt the branch from `rc/5.11` because I am not sure that the merge from rc/5.11 into my branch was made correctly at first.

We should be good now!

73VW: Here's the cypherquery I use:

``` cypher
MATCH (p)
WITH p, rand() AS random
ORDER BY random
LIMIT 1
MATCH path = (p)-[r:links*0..2]-(connected)
RETURN p,path
```

I have multiple objects with `links` (these are relationships with a tag on them). 

The goal is to obtain the graph at a certain depth from a center point (here a random one).

Do you need anything else?

aanastasiou: @73VW Thank you. I am aiming to start looking on this later today or from tomorrow onwards.

All the best
AA

aanastasiou: @73VW Alright, this looks straightforward. 

However, I think that we have a minor clash with capabilities that were added recently.

First of all, please see [here](https://neomodel.readthedocs.io/en/latest/getting_started.html#retrieving-additional-relations) and [here](https://github.com/neo4j-contrib/neomodel/issues/709) for a brief description of that capability. The inner workings of which were largely added [via this](https://github.com/neo4j-contrib/neomodel/pull/686/commits/2ccd44a0d61773c85e204b7f94b145eb9d57064d)

My preference would be to keep only this (#716) contribution for that capability, primarily because it is based on the neo4j driver itself.

The specific example that is presented in the docs would be as simple as:

```
    q = neomodel.db.cypher_query("match p=(a:Person)-[:IS_FROM]->(b:Country) return p", resolve_objects = True)
```

And for an additional example from my testing:

```
    q = db.cypher_query("match p = (a:City)<-[r:LIVES_IN]-(b:Person)-[:IS_FROM]->(c:Country) return p", resolve_objects=True)

This makes `q` a `Path` object that is entirely browsable. And these paths can have any "shape", with any depth, etc.

Now, I know that @mariusconjeaud and @tonioo are already using the 686 pull's changes. Others might be doing so as well.

I would suggest that we have a brief discussion (in any way is convenient) to decide how to proceed with this: 

1. First of all, let's all confirm that it is worth to keep only one implementation of this feature (?) 
2. Secondly, would it be possible to keep the "syntax" of 686 and when the time comes to apply the query to "translate" it to a path? I have not given a lot of thought to this but I can also see the value of making this capability available through neomodel's querying methods, but instead of returning an ad-hoc structure, have it return a Path (?).
3. Finally consider how to proceed with including this feature in an upcoming version. 

@mariusconjeaud and @tonioo For a very minimal "test" of this capability, have a look at [this one](https://gist.github.com/aanastasiou/b89aa67869daaa0c9e135d0128ae086c)

I am aware that a few people might be on vacation around this time, let's try to have this chat as soon as everyone is available.

All the best

aanastasiou: Sorry, forgot to add:

In #716, the Path **needs to apply** object resolution to its nodes and relationships because at the moment it returns them as plain `Node, Relationship` objects. But this is something that is straightforward to add.

73VW: Hey @aanastasiou, I will check that in depth during the week.

Regarding your last comment, it applies object resolution to path sub-objects, check [this](https://github.com/neo4j-contrib/neomodel/pull/716/commits/e7b4975ca7c9c0be87c70418e7e42465aa5332e6#diff-5a14848686f658a9fab17bad447b0d3631c13d078bd3bbde13804138e3150e24R312) and [this](https://github.com/neo4j-contrib/neomodel/pull/716/commits/e7b4975ca7c9c0be87c70418e7e42465aa5332e6#diff-5a14848686f658a9fab17bad447b0d3631c13d078bd3bbde13804138e3150e24R315). 
That makes the resolution recursive. 

Or am I missing something?

aanastasiou: @73VW , the gist I used is [here](https://gist.github.com/aanastasiou/b89aa67869daaa0c9e135d0128ae086c). This was used along with your PR.

When I examine the returned result, I am getting `neo4j.graph.Node` (i.e  `type(q[0][0][0].nodes)`). This behaviour is common at least across Neo4j 4.4.18 and 5.9.

If this is not inline with what you are getting, then please post a detailed setup (neo4j server, neomodel version / commit, etc) along with a gist that demonstrates the bare minimum of the functionality you are contributing. Feel free to use mine as a basis from above. Chances are that this minimal gist will also constitute the basis for writing the pytest tests that test that part of the code. For more information, have a loot in [this directory](https://github.com/neo4j-contrib/neomodel/tree/master/test) (See for example `test_database_management.py`).

aanastasiou: @73VW , any progress with this?

73VW: Hey @aanastasiou, I'm not sure to understant what's wrong with this. Aren't we looking to obtain `neo4j.graph.Node` objects?

I might have missed something here.

Also, I didn't have time to look at the link you pasted. I have given a quick look but couldn't understand everything yet.

mariusconjeaud: Just back from holiday ; waiting for this to be ready then I'll publish version 5.1.1 :-)

aanastasiou: @73VW Path's initialisation does not work as intended. Please see [this](https://github.com/neo4j-contrib/neomodel/pull/738) and more importantly [this](https://github.com/neo4j-contrib/neomodel/pull/738/commits/da411b0780de2f32be744616786f77daade583fb#diff-44a3e8dfb7983646a58a76ce9f70f4251188c28c5206b4369e31fde3b4458675R1).

Issue #717, opened 2023-07-10
TypeError: __new__() got an unexpected keyword argument 'latitude'


raphaelLacerda:
## Expected Behavior (Mandatory)

 I was using neomodel 4.0.8, python 3.9 and this code was working.
```
location = PointProperty(required=False, crs='wgs-84')
def create_location(self, lat, long):
        self.location = NeomodelPoint(latitude=lat, longitude=long)

```

My working pipfile

`
neomodel = { extras = ["extras"], version = "==4.0.8" }
`

## Actual Behavior (Mandatory)

When i upgraded to neomodel 5.0.1, shapely 2.0.1, I got this error

```
>       self.location = NeomodelPoint(latitude=lat, longitude=long)
E       TypeError: __new__() got an unexpected keyword argument 'latitude'
```

My pipfile

```
neomodel = { extras = ["extras"], version = "==5.0.1" }
shapely = "==2.0.1"
```

To run another set of tests, I got the example from the doc 
https://neomodel.readthedocs.io/en/latest/spatial_properties.html#working-with-pointproperty

```
from neomodel import UniqueIdProperty
from neomodel.contrib.spatial_properties import PointProperty, NeomodelPoint
import neomodel

class SomeEntity(neomodel.StructuredNode):
    entity_id = UniqueIdProperty()
    location = PointProperty(crs='wgs-84')

if __name__ == '__main__':
    my_entity = SomeEntity(location=NeomodelPoint((0.0,0.0), crs='wgs-84'))
    print(my_entity.location)
```

Pipfile

```
[packages]
shapely = "==2.0.1"
neomodel = { extras = ["extras"], version = "==5.0.1" }
neo4j = "==4.4.5"

```

And I got the following error:

```
    my_entity = SomeEntity(location=NeomodelPoint((0.0,0.0), crs='wgs-84'))
TypeError: __new__() got an unexpected keyword argument 'crs'

```

Does anyone know how to fix this problem?
If it's a bug related or a compatibility version problem between shapely and neomodel

The code above with this Pipfile works fine

```
shapely = "==1.8.1"
neomodel = { extras = ["extras"], version = "==4.0.10" }
neo4j = "==4.4.5"
```

### Versions

  - OS:
  - Library: neomodel: 5.0.1
  - Neo4j: neo4j:4.4.21

aanastasiou:
[...]

PR #718, opened 2023-07-14
Upgrade `NeomodelPoint` to also work with Shapely 2.0

aanastasiou:
Hello

I have just finished addressing #717 which was an extremely interesting roller coaster ride. It has resulted in two implementations and before we proceed with one of them, I wanted to provide some background:

### The key problem

The folks over at Shapely have changed the way their objects are instantiated across the Python / C divide in a way that affects the way inheritance works when the object is instantiated in Python.

For more information, please see [here](https://github.com/shapely/shapely/issues/1313), [here](https://github.com/shapely/shapely/issues/1233) and [here](https://github.com/shapely/shapely/issues/1698).

Very briefly, certain objects of geometrical models (point, line, polygon and others) must be immutable. The way these are constructed right now, return a fixed and immutable object in Python which cannot be further subclassed.

Neomodel's NeomodelPoint relied heavily on inheritance to offer a NeomodelPoint which would be used for data marshalling to and from Neo4j but also be interoperable with Shapely. That is, once you have a NeomodelPoint result set you could use it as an iterable of Points directly with any other function Shapely offers. No need for the intermediate step of conversion. We already have all the functionality from Shapely and now simply specialise it to allow the creation of specific types of points with the addition of the coordinate reference system (CRS).

### The solution(s)

There are two solutions:

* Insist on inheritance
* Have NeomodelPoint act as an adaptor of Shapely's Point

#### Insist on inheritance

Unfortunately, insisting on inheritance leads to a huge number of hacks. 

For comparison:

The [current implementation](https://github.com/aanastasiou/neomodel/blob/fix_shapely_upgrade/neomodel/contrib/spatial_properties.py#L55) VS [NeomodelPoint trying to inherit from Shapely 2.0's Point](https://github.com/aanastasiou/neomodel/blob/fix_shapely_upgrade/neomodel/contrib/spatial_properties.py#L264)

Very briefly:

1. NeomodelPoint is becoming a "renamed" Shapely Point whose internal structure (i.e. attributes) cannot change (see `__slots__` assignment in `__new__` for example).
2. Its initialisation is now split across `__new__` and ``__init__``
3. To overcome the fact that objects must be immutable, the `_crs` property becomes a dict that maps `id(self)` to the actual
    value of the `_crs` a given object requires.
   * If you do not catch this early, it leads to all sorts of "wonderful" ways by which things can go wrong because this "detail" 
     affects the way `__copy__` and `__deepcopy__` work which are used repeatedly everywhere in Neomodel (e.g. in queries).

The implementation has passed all tests but I would not consider it reliable because you find yourself trying to enforce the way inheritanc should work.

#### Have NeomodelPoint act as an adaptor of Shapely's Point

Compared to insisting on inheritance, this version feels more straightforward, **but** it results in `NeomodelPoint` not being a Shapely Point any more. 

This *might* introduce problems when strict typing is enforced or affect tests where the lineage of the object is taken into account.

This "Adapter" implementation can be found [here](https://github.com/aanastasiou/neomodel/blob/fix_shapely_upgrade_adapter/neomodel/contrib/spatial_properties.py#L265)

Very briefly:

1. The class remains largely untouched, except for getting another attribute to store the underlying Shapely Point which provides the
    actual storage and functions.
2. Two additional functions were added: `__getattr__` to route messages that `NeomodelPoint` does not know how to respond to, to 
    the underlying Shapely Point and `__eq__` to handle comparisons (which before we were getting "for free").
3. The error condition of trying to access "longitude" on a cartesian Point now raises TypeError instead of AttributeError. This change 
    was essential for the properties to operate as expected (if a property raises `AttributeError`, this signals to Python that the property 
    does not exist on this object which then triggers the call to `__getattr__`. 

This implementation also passes all tests right now.

**Given these two pathways**, my personal preference would be to *have NeomodelPoint act as an adapter of Shapely's Point*.

If however in the future, [the folks at Shapely decide to change their mind about the way Shapely objects are instantiated](https://github.com/shapely/shapely/issues/1698#issuecomment-1634802367), I would happily change the implementation back to NeomodelPoint being a child of Shapely's Point.


aanastasiou:
@mariusconjeaud We might have to add another build conf for Shapely < 2.0

mariusconjeaud:
> @mariusconjeaud We might have to add another build conf for Shapely >= 2.0

Not sure what you mean by this ?

FYI, the tests pipeline failed here, but only because it didn't manage to connect to Aura in the Aura-specific tests, which sometimes happens since I use a test Aura instance which looks like it's not always on.

aanastasiou:
@mariusconjeaud 

I understand. The way the package stands right now, [specifies Shapely < 2.0](https://github.com/neo4j-contrib/neomodel/blob/master/pyproject.toml#L55).

This PR supports both. It detects the version and establishes `NeomodelPoint` accordingly.

I did this because I did not want to outright deprecate Shapely < 2.0. Users now have both options. If the transition to Shapely 2.0 creates any problems, they can report them and we can fix them but they also get the option to work with earlier Shapely versions.

PR #720, opened 2023-07-17
Fix script invocation

aanastasiou:
Installing via `pyproject.toml` has brought about some subtle changes compared to the way things were installed before. 

The way `pyproject.toml` is setup to "read" modules, makes it recognise `scripts` as a separate (to `neomodel`) package and install it as such. 

Consequently, you can `import scripts` (no relationship to `neomodel`) and at the same time, `docker-neo4j.sh` is not provided at the command line anymore (which is alright).

Recognising scripts as a separate module is problematic from a code organisation point of view but can cause more serious problems if it "collides" with an existing module `scripts`.

This PR fixes that by making `scripts` part of the `neomodel.module`.

The user scripts `neomodel_install_labels`, `neomodel_remove_labels` are accessible in exactly the same way as before (absolutely **no** change for the end-user) without changes to `pyproject.toml`


mariusconjeaud:
I will fix the Integration tests GH Actions pipeline that cannot find the Neo4j docker creation script anymore.

mariusconjeaud:
Thanks for fixing the pipeline ! One conflict left and then we can merge this (so that it's there in next release)

mariusconjeaud (approved):

PR #723, opened 2023-07-19
5.1.0 release

mariusconjeaud:
Version 5.1.0 2023-07
* Bumped neo4j-driver version to 5.10.0
* Breaking change : When using neomodel along with Neo4j version 5, use StructuredNode and StructuredRel's element_id property instead of id. If you have Cypher queries which currently use the id() function, migrate them to elementId() instead.
If you use Neo4j version 4.4, this change should be transparent to you, and you should keep using id() function in Cypher since elementId() is not implemented in 4.4.
* Bump Shapely to >= 2.0.0
* Fix schema scripts invocation
* Removed custom __str__ method, part of NodeClassRegistry, from Database() (db) object.


neomodel/match.py
```diff
@@ -818,9 +948,7 @@ def __init__(self, source, name, definition):
             "relation_type",
         }
         if invalid_keys:
-            raise ValueError(
-                f"Unallowed keys in Traversal definition: {invalid_keys}"
```
aanastasiou, line 822: A better way to put this would be "Prohibited keys...."

PR #730, opened 2023-08-01
Updated ``neomodel.db.set_connection``

aanastasiou:
At its current state, ``neomodel.db.set_connection`` attempts to populate ``db.database_version`` by making a call to the database server. This assumes that at the time of calling ``set_connection`` the database server is already up and running.

This is not always the case and so far, the purpose of ``set_connection`` has been just that, to just set the connection, not to try and perform any operations on the database.

In any case, as useful as ``database_version`` is, it is not required all the time. Code that needs ``database_version`` will call it explicitly. 

This PR attempts to fix this in the most straightforward way possible: ``database_version`` gets updated when required and that information is cached for future calls.

Another possibility is to automatically try to update ``database_version`` upon ``ensure_connection`` but at the moment I went for the most straightfowrard option.


mariusconjeaud:
The database version is only needed to choose between the id() and elementId() Cypher methods indeed, so I think this is a good catch.

I resolved the merge conflicts and we can merge this in for release 5.1.1 then. 

Thanks !

PR #738, opened 2023-08-22
Automatic path inflation

aanastasiou:
Addresses shortcomings of [716](https://github.com/neo4j-contrib/neomodel/pull/716).

* Paths are properly instantiated
* Adds tests and documentation

Path failed to instantiate properly, [on this line](https://github.com/neo4j-contrib/neomodel/pull/716/commits/e7b4975ca7c9c0be87c70418e7e42465aa5332e6#diff-5a14848686f658a9fab17bad447b0d3631c13d078bd3bbde13804138e3150e24R315), when relationships that were not association classes were passed to it. This is because, those relationships do not have data and therefore do not have an entry in ``_NODE_CLASS_REGISTRY``. That lookup failed (unfortunately silently) and resulted in a ``Path`` object with "unresolved" ``nodes, relationships``. This was corrected to produce a ``StructuredRel`` in that case. All objects (whether relationship or node) load exactly once.


mariusconjeaud:
Adds the following test fail : 
image

mariusconjeaud:
Merged the conflict because it was my fault. I accidentally removed the logger in rc branch when I removed the "neo4j" log watcher

mariusconjeaud (approved):

PR #760, opened 2023-10-26
Add pandas and numpy format support

mariusconjeaud:
Adds support for casting cypher_query results to :

- pandas DataFrame and Series
- numpy Array


aanastasiou:
Are you already working on this? I have done this already in a different setting

mariusconjeaud:
@aanastasiou yes it's done already, but it took only a few minutes so I'm OK destroying it if you have a nicer way !

Issue #769, opened 2023-12-08
Incorrect query for count of objects


OlegGustilov:
## Expected Behavior (Mandatory)
The method _count() from[match.py]( https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/match.py#L660) should consider case to properly build the query with "skip" and "limit" clauses.

## Actual Behavior (Mandatory)
The method _count() may build a non-working query with clauses "skip" and "limit". As result the Neo4J returns incorrect or empty (with "skip" clause) results.

## How to Reproduce the Problem
In my case I am using Django framework and Django admin functionality.
I created simple model and simple admin view to manage my objects for example Users list.
When I trying to click next page using pagination in the Django Admin UI I see error "index out of range". 
It happens when Django trying to get count of items on the page, as result the method above will be called.

The same can happen in different cases without Django.

### Simple Example

Datasets and Statements
The method "_count()" may build following Neo4J query:
```
MATCH (user:User) RETURN count(user) SKIP 100 LIMIT 100
```
Such query returns empty results.

## Screenshots (where it's possibile)
-

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: 
  - Library: neomodel 5.2.0
  - Neo4j: 4.4.19

mariusconjeaud:
Hello,

Thanks for this interesting feedback.

So, the first idea that comes to mind (dropping SKIP and LIMIT when counting) is probably not the right solution, since your use case shows that we might still want to count items even when using skip and limit (for pagination purposes).

So the problem here is rather that instead of this query :
`MATCH (user:CTTermRoot) RETURN count(user) SKIP 100 LIMIT 100 => Returns nothing`

It should generate this query :
`MATCH (user:CTTermRoot) WITH user SKIP 100 LIMIT 100 RETURN count(user) => Returns number of elements in your page`

mariusconjeaud:
Please review the PR : #771

PR #771, opened 2023-12-11
Incorrect query for count of objects #769

mariusconjeaud:
Issue was when count was used in combination with skip/limit.

Had to change the query builder so that pagination happens to variables before the count aggregation is applied.


OlegGustilov:
Thank you for the fix! 
I tested it with Django and it works good!

PR #777, opened 2024-01-12
Add wraps to wrapper defined in TransactionProxy

giosava94:
Hi all,

first for all, thanks for your work and for keeping up to date this library!

I'm using `neomodel` in a FastAPI project to update a neo4j database. For each endpoint reading or writing to the database, I open a new DB transaction using the decorators `write_transaction` and `read_transaction`. 

Since I'm using multiple wrappers from different libraries for each endpoint, I noticed that the `TransactionProxy` `__init__` and `__call__` functions were called when creating the endpoint, but the `wrapper` defined in `__call__` and the `__enter__` and `__exit__` functions were not invoked.

This pull request adds the `functools` `wraps` function to that `wrapper`. Doing this I resolved the problem described before. 

I executed the tests with pytest and they passed. I also noticed that the behavior with easier use cases does not change.

Let me know if this change introduces breaking changes or strange behavior (hopefully not).

Regards

### DIFF
--- a/neomodel/util.py
+++ b/neomodel/util.py
@@ -3,6 +3,7 @@ import os
 import sys
 import time
 import warnings
+from functools import wraps
 from threading import local
 from typing import Optional, Sequence
 from urllib.parse import quote, unquote, urlparse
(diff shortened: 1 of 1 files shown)


mariusconjeaud:
Tested with the new async code (merge in rc/5.3.0) and it works just fine.

So I will merge this into a temp branch (rc/5.3.0-pre), so I can merge it back into the actual release candidate branch. This is because I otherwise need to just copy paste your code as the async and sync codebases are now split, and you won't be seen as a contributor anywhere :-)

Issue #778, opened 2024-01-12
Structured Node `__eq__` works incorrectly for non-persisted nodes. 


OlehChyhyryn:
## Expected Behavior (Mandatory)

If nodes are not created in the database and do not have an `element_id` property, they should be compared by `id` in memory or always count as different. 

## Actual Behavior (Mandatory)

All non-persisted node countries as equal

## How to Reproduce the Problem

Create any StructuredNode objects without saving them to the database. Call comparison. 

### Simple Example

```
from neomodel import StructuredNode

class Foo(StructuredNode):
    pass

a = Foo()
b = Foo()

print(a == b)
```

The problem comes from the `StructuredNode.__eq__` function.

```
    def __eq__(self, other):
        if not isinstance(other, (StructuredNode,)):
            return False
        if hasattr(self, "element_id") and hasattr(other, "element_id"):
            return self.element_id == other.element_id
        return False
```

I haven't prepared a pull request because I'm unsure which behavior should be selected as a default: all non-persisted nodes. Are they not equal, or do we want to compare them by memory ID or the other way around? 

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: WSL Ubuntu 22.04 + Windows 11
  - Library: 5.2.1
  - Neo4j: 5.16.0

PR #780, opened 2024-01-15
fix(core): compare objects by id in memory if they are different

OlehChyhyryn:
github: [https://github.com/neo4j-contrib/neomodel/issues/778](Structured Node __eq__ works incorrectly for non-persisted nodes.)


OlehChyhyryn:
Something weird with tests

PR #782, opened 2024-01-30
Added random postfix to the ref variable, to avoid reserve words collision

herrjemand:
The issue I had is that my Node class was name `Case`. When label is generated it is using lower case class name, so it was using `case` as variable

```
MATCH (case) WHERE elementId(case)=$case_1 WITH case MATCH (case)-[r1:`PARTICIPATES`]-(participants_r1:Applicant) RETURN participants_r1
```

that line

`MATCH (case)` is interpreted by CQL as MATCH (CASE).

So I've added random prefix to the variables and labels to prevent accidental reserve words panics in future

This is a result query:

```
MATCH (case_3023) WHERE elementId(case_3023)=$case_3023_1 WITH case_3023 MATCH (case_3023)-[r1:`PARTICIPATES`]-(participants_r1:Applicant) RETURN participants_r1
```

### DIFF
--- a/neomodel/match.py
+++ b/neomodel/match.py
@@ -1,5 +1,6 @@
 import inspect
 import re
+import secrets
 from collections import defaultdict
 from dataclasses import dataclass
 from typing import Optional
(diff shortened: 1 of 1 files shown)


mariusconjeaud:
This breaks tests, and especially breaks the fetch_relations with filter() applied (see test_match_api.py::test_fetch_relations)

PR #783, opened 2024-02-05
Add async support to neomodel

mariusconjeaud:
Introducing support for asynchronous programming to neomodel required to introduce some breaking changes:
- Replace `from neomodel import db` with `from neomodel.sync_.core import db`
- config.AUTO_INSTALL_LABELS has been removed. Please use the `neomodel_install_labels` command instead.


mariusconjeaud:
!!!!!!! The integration tests pipeline was fully successful for the first time since starting the work on async !!!!!!!

It should be ready soon !

PR #829, opened 2024-09-20
Added support for annotations and calling aggregating functions.

tonioo:
Also introduced a new traverse_relations() method which only return the first hop of the path in the return clause.


mariusconjeaud:
Will add doc later - more features coming in this featureset

PR #832, opened 2024-10-02
Subqueries support

tonioo:
Simple way to include one or more subqueries within a NodeSet context


mariusconjeaud:
Example use case : insert an isolated CALL {} subquery to retrieve the first element of an ordered collection of relationships

Issue #834, opened 2024-10-13
Allow to store UTF-8 values in JSONProperty


danikirish:
## Feature description (Mandatory)
Storing non-ASCII characters in `JSONProperty` will turn them into escaped Unicode characters, due to default kwarg `ensure_ascii=True` of `json.dumps()`. 
This feature would allow passing the `ensure_ascii` kwarg to `JSONProperty` to control this behaviour. By setting the default value as `True`, this would ensure consistency with existing usage of this property.

#### Example before
```python
class Item(StructuredNode):
    name = StringProperty()
    extra_info = JSONProperty(default={})

extra_info = {"name_ja": "水"}
item = Item(name="water", extra_info=extra_info).save()
# extra_info would be saved as '{"name_ja": "\\u6c34"}' in the database.
``` 

#### Example after
```python
class Item(StructuredNode):
    name = StringProperty()
    extra_info = JSONProperty(default={})
    extra_info_unicode = JSONProperty(default={}, ensure_ascii=False)

extra_info = {"name_ja": "水"}
item = Item(name="water", extra_info=extra_info, extra_info_unicode=extra_info).save()
# extra_info would be saved as '{"name_ja": "\\u6c34"}' in the database.
# extra_info_unicode would be saved as '{"name": "水"}' in the database.
```

## Considered alternatives
Create a separate `UnicodeJSONProperty` property class.

## How this feature can improve the project?
- Makes neomodel's behaviour more consistent with neo4j, since it respects UTF-8 encoding
- Improves readability
- Improves data integrity by keeping the stored value same as the original data

danikirish:
Pull request: https://github.com/neo4j-contrib/neomodel/pull/835

PR #835, opened 2024-10-13
Add UTF-8 support to JSONProperty

danikirish:
Implements #834


mariusconjeaud:
Thank you !

Could you add a test for this ?

PR #875, opened 2025-06-11
Mutually exclusive relationships #871

### LINKED ISSUE #871: Constraint between relationships (OR operator)
## Feature description

It woulde be nice to add a constraint definition for cases where a node could have one or the other relationship (similar to the OR operator), but not both.

For example, consider the following (dummy) case: a _Person_ can only have one type of pet, either a _Dog_ or a _Cat_.

```python
class Dog(StructuredNode):
    name= StringProperty(required=True)

class Cat(StructuredNode):
    name= StringProperty(required=True)

class Person(StructuredNode):
    name= StringProperty(required=True)

    cat= RelationshipTo('Cat', 'HAS_PET', cardinality=ZeroOrMore)
    dog= RelationshipTo('Dog', 'HAS_PET', cardinality=ZeroOrMore)
```

This is, of course, a simplified example to illustrate the required constraint, without the need for creating an _Animal_ superclass.

mariusconjeaud:
closes #871


mariusconjeaud:
Replace by PR #963

Issue #872, opened 2025-05-18
Query Vector Indexes


greengori11a:
## Feature description 

neomodel should support querying vector indexes using the neo4j procedures [`db.index.vector.queryNodes`](https://neo4j.com/docs/cypher-manual/current/indexes/semantic-indexes/vector-indexes/#query-vector-index) and [`db.index.vector.queryRelationships`](https://neo4j.com/docs/cypher-manual/current/indexes/semantic-indexes/vector-indexes/#query-vector-index).  #815 added support for creating vector indexes but querying does not appear to be currently supported. 

## How this feature can improve the project?

This would improve support for semantic ML/AI workflows in neo4j and neomodel.

greengori11a:
I am happy to have a stab at this, any suggestions on implementation?

aanastasiou:
@greengori11a Just to clarify, you still have the option to send the query using [`cypher_query`](https://neomodel.readthedocs.io/en/latest/cypher.html).

There are some considerations about its use: The most straightforward example is queries that have both a vector and property part. In general, you would aim to first do the property based filtering of the nodes followed by the vector "filter". In other words, first query nodes by the desired property values and then order the result of that query by its similarity to a given vector. But that doesn't mean that the reverse might not be desireable in some cases.

In terms of adding something like this to neomodel:

Neomodel provides a composable interface to writing CYPHER queries. This basically means that when you write something like [`SomeEntity.nodes.filter(some_attribute__gt=10`](https://neomodel.readthedocs.io/en/latest/getting_started.html#retrieving-nodes), neomodel builds up a cypher query sends it to the database (with `cypher_query` by the way) and then handles the data transformation for the returned results. The starting point for that is [`.nodes`](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/sync_/core.py#L1508) and the "composition" is possible via [`NodeSet`](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/sync_/match.py#L1355). For more details please see [`match`](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/sync_/match.py) and ['match_q'](https://github.com/neo4j-contrib/neomodel/blob/master/neomodel/match_q.py) (but notice that `match` has both sync and async versions).

[...]

PR #885, opened 2025-07-22
Implementing #872 - Query Vector Indexes

greengori11a:
Apologies for the slow development time on this feature. 

This PR implements Querying Vector Indexes #872 on nodes using the `.filter()` function flow. 

I've written some testing for the VectorFilter querying, but am unsure if the coverage of the testing is enough (it probably isnt, if you have any guidance on other things to test I am happy to implement them). 

I've also written some documentation about how to define Vector and Full Text Indexes as I think it was missing from the documentation. 

There are some weird files changed within this PR also, this is as a result of running black formatting.

### DIFF
--- a/test/async_/test_vectorfilter.py
+++ b/test/async_/test_vectorfilter.py
@@ -1,43 +1,21 @@
-from test._async_compat import mark_async_test
-from test.conftest import check_and_skip_neo4j_least_version
 from datetime import datetime
-from neomodel.semantic_filters import VectorFilter
+from test._async_compat import mark_async_test
+
+import pytest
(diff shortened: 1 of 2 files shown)


greengori11a:
Hi @mariusconjeaud, 

I have updated the tests to skip for Neo4j versions that do not have VectorIndexes implemented. 
I have added a test that shows the VectorFilter works with other django filters, are there any specific filters / advanced query methods that you would like me to test for? I will look to test relationship traversals filtering as well. 
I have made some updates to the documentation to reduce duplication. 

Thanks

Issue #888, opened 2025-08-25
Concurrency issues with multiple async contexts


EricAtORS:
## Expected Behavior (Mandatory)
When entering multiple async transactions, they should all run independently and be allowed to run.

## Actual Behavior (Mandatory)
I get this error:
```
  File "/project/src/backend_api_server/webserverapi/routers/projectitem/common.py", line 395, in complete_multipart_upload
    async with adb.transaction:
               ^^^^^^^^^^^^^^^
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neomodel/async_/core.py", line 1290, in __aexit__
    await self.db.rollback()
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neomodel/async_/core.py", line 81, in wrapper
    return await func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neomodel/async_/core.py", line 330, in rollback
    await self._session.close()
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neo4j/_async/work/session.py", line 222, in close
    await self._connection.fetch_all()
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neo4j/_async/io/_bolt.py", line 879, in fetch_all
    detail_delta, summary_delta = await self.fetch_message()
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neo4j/_async/io/_bolt.py", line 861, in fetch_message
    tag, fields = await self.inbox.pop(
                  ^^^^^^^^^^^^^^^^^^^^^
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neo4j/_async/io/_common.py", line 77, in pop
    await self._buffer_one_chunk()
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neo4j/_async/io/_common.py", line 54, in _buffer_one_chunk
    await receive_into_buffer(self._socket, self._buffer, 2)
  File "/project/src/backend_api_server/.venv/lib/python3.12/site-packages/neo4j/_async/io/_common.py", line 345, in receive_into_buffer
    n = await sock.recv_into(
        ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #889, opened 2025-08-26
Fix for AsyncDatabase and parallel transactions #888 ?

DenesPal:
async_.core.AsyncDatabase was incorrectly subclassed from threading.local, that did not make it's properties context-local, leaking state (like session and transaction states) across async contexts.

I am assuming that sync_.core.Database implementation is correct.
Apart from the two class-level global mappings (_NODE_CLASS_REGISTRY and _DB_SPECIFIC_CLASS_REGISTRY), all instance properties of Database are threading-local.
So I've converted the same instance properties of AsyncDatabase to ContextVar @property-s (because ContextVars work with .get & .set).


mariusconjeaud (approved):

mariusconjeaud:
OK tests are fixed, it was an issue with a previous PR that had duplicated code which I had missed

Issue #464, opened 2019-08-02
Unreasonable cardinality constraints


mfboulos:
Hello all,

Before getting into this, I read the docs, and understand that the current capabilities surrounding the `cardinality` of relationships entails:

> It is possible to (softly) enforce cardinality constraints on your relationships. Remember this needs to be declared on both sides of the relationship definition

I'm assuming that what is meant by "softly" is that this cardinality is only enforced on the connecting side of the relationship.

Take for example, my use case. I have a single node that has `ZeroOrMore` `OUTGOING` relationships, each to nodes that have `One` `INCOMING` relationship corresponding to that. Kinda looks like this:

![image](https://user-images.githubusercontent.com/25018328/62335792-427e1300-b49b-11e9-8d22-0a05202092f1.png)

Essentially, the `One` constraint should make this following case impossible:
![image](https://user-images.githubusercontent.com/25018328/62336031-5f671600-b49c-11e9-8a19-28eed5bd69ec.png)

Which it is, if I try to connect it from the `StructuredNode` that has the `RelationshipDefinition` constrained by `One`. But if I connect from the `ZeroOrMore` connected node, the other node's constraints aren't checked. Example code below:

```python
class Test(StructuredNode):
    a = StringProperty(unique_index=True)
    other = RelationshipFrom('Test2', 'IS_UNDER', cardinality=One)

    def __init__(self, a, *args, **kwargs):
        kwargs['a'] = a
        super().__init__(*args, **kwargs)

class Test2(StructuredNode):
    b = StringProperty(unique_index=True)
    other = RelationshipTo('Test', 'IS_UNDER', cardinality=ZeroOrMore)

    def __init__(self, b, *args, **kwargs):
        kwargs['b'] = b
        super().__init__(*args, **kwargs)

install_all_labels()

testHub = Test2('a').save()
single1 = Test('b').save()
single2 = Test('c').save()
single3 = Test('d').save()

testHub2 = Test2('z').save()

testHub.other.connect(single1)
testHub.other.connect(single2)
testHub.other.connect(single3)

testHub2.other.connect(single1) # should fail, but doesn't
single1.other.connect(testHub2) # should fail, and does
```

Should the target of a relationship be able to establish connections? And either way, shouldn't both source and target be responsible for understanding the constraints of each side? This feels like a pretty common use case to me, so I'm curious to see if anyone else shares this concern.

aanastasiou:
[...]

Issue #891, opened 2025-08-29
RelationshipFrom cardinality constraints are not enforced, unlike RelationshipTo


billycalladine:
### Description

The `RelationshipFrom` relationship manager does not enforce cardinality constraints when they are violated, while `RelationshipTo` correctly throws an error under the same conditions. This creates an inconsistency in the relationship validation behavior and can lead to data integrity issues.

### Current behaviour

When using `RelationshipFrom` with cardinality constraints (e.g., `cardinality=One`, `cardinality=ZeroOrOne`), violations of these constraints do not raise any errors and the operations complete successfully, potentially creating invalid relationship structures in the database.

### Expected behaviour

`RelationshipFrom` should enforce cardinality constraints consistently with `RelationshipTo`, throwing an appropriate error (e.g., `AttemptedCardinalityViolation`) when the specified cardinality would be violated.

### Steps to reproduce

```python
from neomodel import StructuredNode, RelationshipTo, RelationshipFrom, StringProperty, One

class Person(StructuredNode):
    name = StringProperty()
    # This enforces cardinality correctly
    has_single_car = RelationshipTo('Car', 'OWNS', cardinality=One)
    
class Car(StructuredNode):
    model = StringProperty()
    # This does NOT enforce cardinality
    owned_by = RelationshipFrom('Person', 'OWNS', cardinality=One)

# Create instances
person1 = Person(name="John").save()
person2 = Person(name="Jane").save()
car = Car(model="Tesla").save()

# This works (first connection)
person1.has_single_car.connect(car)

# Attempting to connect from another person to the same car
# Expected: Should raise AttemptedCardinalityViolation
# Actual: Succeeds without error
person2.has_single_car.connect(car)  # This would violate the RelationshipFrom cardinality=One on Car
```

### Environment

- **neomodel version**: 5.4.1
- **Neo4j version**: 5.26.0
- **Python version**: 3.10
- **Operating System**: MacOS 15.6.1 

### Additional context

The [docs](https://neomodel.readthedocs.io/en/latest/relationships.html#cardinality) state:

> It is possible to (softly) enforce cardinality constraints on your relationships. Remember this needs to be declared on both sides of the relationship definition
> 
> If a cardinality constraint is violated by existing data a **CardinalityViolation** exception is raised.
> 
> On attempting to violate a cardinality constraint a **AttemptedCardinalityViolation** is raised.

### Notable info

This issue appears to be related to previously reported issues:
[...]

Issue #867, opened 2025-03-15
Cardinality on both side of pairs


fsalmasri:
in the example: 

```
class Person(StructuredNode):
    car = RelationshipTo('Car', 'CAR', cardinality=ZeroOrMore)


class Car(StructuredNode):
    owner = RelationshipFrom('Person', 'CAR', cardinality=One)
```

The relationship cardinality is different on both sides. Person can have zero or more cars, but car can be owned by one person. 

when connecting two persons two car this won't raise an error, but when connecting car two persons this will raise an error. cardinality is only checked on the car side. 

why cardinality is not verified, on both pairs when relationship is connected. 

Also car nodes can't be created without a connection without raising an error even though the cardinality is One.

PR #892, opened 2025-08-29
Fix: Enforce cardinality on RelationshipFrom and bidirectional relationships

billycalladine:
### Context

This pull request addresses a critical data integrity issue where cardinality constraints on `RelationshipFrom` were not being enforced, leading to potential inconsistencies in the graph data. This change introduces bidirectional cardinality validation to ensure that relationship constraints are respected from both the source and target nodes' perspectives.

### Problem

Previously, neomodel only validated the cardinality of a relationship from the source side (the node on which the `.connect()` method is called). This meant that a `RelationshipTo` with a specific cardinality (`One`, `ZeroOrOne`) would be correctly validated, but the corresponding `RelationshipFrom` on the target node would not be.

For example, if a `Pet` could only have one `Owner` (`RelationshipFrom('Owner', 'OWNS', cardinality=One)`), it was still possible to connect multiple `Owner` nodes to the same `Pet`, violating the constraint without raising an error.

This fixes issues #464 and #867 and [#891](https://github.com/neo4j-contrib/neomodel/issues/891)

### Solution

The core of the solution is to enforce cardinality checks on both ends of a relationship during the `connect` operation.

1.  **Bidirectional Validation in `connect`**:
    *   The `connect` method in `RelationshipManager` has been updated to perform two checks:
        1.  **Local Cardinality**: It validates the cardinality on the source relationship, as before.
        2.  **Remote Cardinality**: It now inspects the target node to find the inverse relationship definition. If found, it validates the cardinality from the target node's perspective *before* creating the relationship.

2.  **`_check_cardinality` Method**:
    *   A new method, `_check_cardinality`, has been introduced to the base `RelationshipManager`.
    *   This method is implemented in the specific cardinality classes (`One`, `ZeroOrOne`) to encapsulate the validation logic.
    *   The original validation code was moved from the `connect` methods of the cardinality classes into this new method, centralising the logic.

### How to verify

A test case has been added to confirm the fix. The following scenario now works as expected:

```python
class Owner(StructuredNode):
    name = StringProperty()
    pets = RelationshipTo('Pet', 'OWNS')

class Pet(StructuredNode):
    name = StringProperty()
    owner = RelationshipFrom('Owner', 'OWNS', cardinality=One)

owner1 = Owner(name="Alice").save()
owner2 = Owner(name="Bob").save()
pet = Pet(name="Fluffy").save()

# This connection succeeds
owner1.pets.connect(pet)

# This connection now correctly fails with AttemptedCardinalityViolation
with pytest.raises(AttemptedCardinalityViolation):
    owner2.pets.connect(pet)
```

### Impact

*   **Data Integrity**: This change significantly improves data integrity by preventing invalid relationship structures.
*   **Backward Compatibility**: This is a breaking change for applications that implicitly relied on the buggy behaviour of unenforced `RelationshipFrom` cardinality. Code that created relationships violating these constraints will now raise an `AttemptedCardinalityViolation`. This is the correct and expected behaviour.
*   **Performance**: The fix introduces a small overhead by requiring an additional lookup for the inverse relationship definition on the target node. This is a necessary trade-off for ensuring data correctness and is performed in-memory without extra database queries.


mariusconjeaud (approved):

mariusconjeaud:
Hello !

Thanks a ton for this very well documented and explained Issue & PR ! And also for the thorough research and linking with similar issues.

I have been away from neomodel issues for the past few months, so apologies for the delay in reviewing this.
I will approved and merge for the next minor release, which would happen very soon !

billycalladine:
@mariusconjeaud thanks for the approval and merge! 🥳

On reflection, this could be a breaking change for unsuspecting users that are using `RelationshipFrom` but without the knowledge that it isn't currently working as expected. Having said that, if you are specifying cardinality in your code then you should expect this to work. So while this PR fixes cardinality not being enforced for `RelationshipFrom`, it may introduce breaking behaviour for existing users.

Specifically, applications that currently connect multiple nodes through a `RelationshipFrom` with `cardinality=One` or `ZeroOrOne` will start raising `AttemptedCardinalityViolation`. From a correctness perspective, this is the expected behaviour (and it matches the documentation as well as `RelationshipTo`). However, in practice, users who were unaware of the bug may experience errors after upgrading.

Nevertheless, since developers explicitly declare cardinality in their model, it’s reasonable they should expect enforcement. The fix makes the API consistent and aligns with the documented contract, but I wanted to flag the behavioural change so it’s not surprising to users.

billycalladine:
my bad, I have just caught up with everything and read your comment [here](https://github.com/neo4j-contrib/neomodel/issues/896). You're already way ahead of me 😅 

Thanks for resolving 🙌

PR #907, opened 2025-10-05
Implement FulltextFilter

greengori11a:
Implement FullTextIndexFilter
Add documentation for FullTextIndexFilter
Little cleanup on Semantic Index documentation

Apologies for the new PR - I think i messed up my rebase so badly that it was worth restarting. 
Let me know what you think.

### DIFF
--- a/doc/source/semantic_indexes.rst
+++ b/doc/source/semantic_indexes.rst
@@ -6,7 +6,7 @@ Semantic Indexes
 
 Full Text Index
 ----------------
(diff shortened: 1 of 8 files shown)


mariusconjeaud:
Thank you ! Will review. Set the target to the rc/6.0 branch because I hope to make a major release next and no minor ones in between, but this might change.

I can fix the merge conflicts if we do want to get this for release 6.0 no worries (it's mostly type hinting as the next major will drop support for Python 3.8 and can use newer type hints)

PR #910, opened 2025-10-06
issue/909 - improve error message verbosity

billycalladine:


mariusconjeaud:
Thank you for this ! I tried to unify the error messages, and maybe was a bit too extreme. :-)

Setting it for the rc 6.0 branch because I hope it's the next release ; will change if relevant.

Issue #826, opened 2024-08-22
NodeClassAlreadyDefined Exception in Streamlit Application due to Repeated Class Registration


muelldlr:
## Expected Behavior (Mandatory)
When repeatedly reloading a Streamlit application, classes should not be registered multiple times in the neomodel library in order to avoid a `NodeClassAlreadyDefined` exception. The application should be executed without errors, even if it is reloaded several times.

## Actual Behavior (Mandatory)
When the Streamlit page is reloaded, a `NodeClassAlreadyDefined` exception occurs because the classes are registered again. This leads to an error and prevents the application from continuing to run.

## How to Reproduce the Problem
Create a simple Streamlit application that uses neomodel classes.
Reload the Streamlit page several times.
Observe that a NodeClassAlreadyDefined exception is thrown on each reload.

### Simple Example

```
import streamlit as st
from neomodel import StructuredNode, StringProperty, UniqueIdProperty, config

config.DATABASE_URL = 'bolt://neo4j:neo4j@localhost:7687'

class Component(StructuredNode):
    uid = UniqueIdProperty()
    name = StringProperty(unique_index=True)
    version = StringProperty()

st.write("This is a test application")
```

## Screenshots (where it's possibile)
![grafik](https://github.com/user-attachments/assets/441742d7-cf0d-4c41-abe0-e68d2a895266)

## Specifications (Mandatory)

Currently used versions

### Versions

  - OS: Microsoft Windows 10
  - Library:  5.3.2
  - Neo4j: 5.22.0

miguelfg:
Same here using Django running `python manage.py runserver` also with any of its parameters ` --noreload --nothreading `

mruge-shr:
with a little help from GPT4 i found this workaround that seems to work on mine

```python
from neomodel.sync_.core import db  # Import the Neo4j Database singleton instance

def get_or_create_entity_node_class():
    """
    Retrieve the EntityNode class from the registry or define it if not present.
    """
    # Check if 'EntityNode' is already registered in the registry
    registry = db._NODE_CLASS_REGISTRY  # Access the internal registry
    for labels, node_class in registry.items():
        if 'EntityNode' in labels:
            print("EntityNode found in registry.")
            return node_class

    # If not found, define and register the EntityNode class
    class EntityNode(StructuredNode):
        """Top-level Neo4j node class for storing UUIDs and relationships."""
        uuid = UniqueIdProperty()
[...]

PR #923, opened 2025-11-28
#826 Allow node definition reload

mariusconjeaud:
Node class definitions now accept a `__allow_reload__` flag ; if set to True, then trying to redefine the class will only raise a warning, not an error. 

This is useful for hot reloading purposes like streamlit or Django applications


mariusconjeaud:
EDIT : Moved allow_reload to global config, so it becomes a real development mode parameter, enabled everywhere. Defaults to False

PR #936, opened 2026-04-23
Added transaction timeout support

DenesPal:
## Summary

- Add transaction_timeout configuration option (seconds, default None) that sets a server-enforced timeout on all
transactions - both explicit `db.transaction` / `db.begin()` and auto-commit `db.cypher_query()` outside a transaction
- Support per-transaction timeout override via `db.begin(timeout=...)` or `AsyncTransactionProxy(db, timeout=...)`
- Configurable via `get_config().transaction_timeout`, the legacy `config.TRANSACTION_TIMEOUT`,
  or the `NEOMODEL_TRANSACTION_TIMEOUT` environment variable

## Motivation

Neomodel provided no way to limit how long a transaction could run.
The Neo4j driver already supports per-transaction timeouts via
`begin_transaction(timeout=...)` and `Query(text, timeout=...)`,
but neomodel never passed these through.

E.g. in a REST API environment, transaction timeout should be lower than the HTTP gateway timeout,
so clients can always get the result of the operation, and back-end does not get choked with never-ending slow queries.
(Can't just rely on the back-end e.g. FastAPI to detect the client disconnect in time and kill the db transaction).

### neomodel/config.py

- New transaction_timeout: `float | None` field with env var `NEOMODEL_TRANSACTION_TIMEOUT`
- Validation (must be positive if set), `from_env` handling for `float | None`, legacy `TRANSACTION_TIMEOUT` property

### neomodel/async_/database.py
(sync code generated via `make-unasync`)

- `begin()` new *timeout* parameter, falls back to `config.transaction_timeout`, forwarded to
  `session.begin_transaction(timeout=...)`
- `cypher_query()` ad-hoc queries wrap the query string in `neo4j.Query(query, timeout=...)` when config timeout is set
- `_run_cypher_query()` type hint to accept `str | Query`

### neomodel/async_/transaction.py
(sync code generated via `make-unasync`)

- `AsyncTransactionProxy.__init__` new *timeout* parameter, forwarded to `db.begin()`

## Tests added

- `test_transaction_timeout_default` verifies timeout=None passed by default
- `test_transaction_timeout_explicit` verifies explicit db.begin(timeout=7.5)
- `test_transaction_timeout_fires` 1s timeout to AsyncTransactionProxy with 3s db sleep raises ClientError
- `test_transaction_timeout_via_config_fires` config-level timeout with 3s db sleep raises ClientError
- `test_transaction_timeout_succeeds` 2s timeout with 0.5s sleep completes normally
- `test_cypher_query_transaction_timeout_via_config_fires`
  cypher_query with 1s config-level timeout with 3s db sleep raises ClientError
- `test_cypher_query_transaction_timeout_via_config_succeeds`
  cypher_query with 2s config timeout and 0.5s db sleep completes normally

### DIFF
--- a/neomodel/async_/database.py
+++ b/neomodel/async_/database.py
@@ -17,6 +17,7 @@ from neo4j import (
     AsyncResult,
     AsyncSession,
     AsyncTransaction,
+    Query,
     basic_auth,
 )
 from neo4j.api import Bookmarks
@@ -441,7 +442,10 @@ class AsyncDatabase:
 
     @ensure_connection
     async def begin(
-        self, access_mode: str = ACCESS_MODE_WRITE, **parameters: Any
+        self,
+        access_mode: str = ACCESS_MODE_WRITE,
+        timeout: float | None = None,
+        **parameters: Any,
     ) -> None:
         """
         Begins a new transaction. Raises SystemError if a transaction is already active.
@@ -462,7 +466,10 @@ class AsyncDatabase:
         )
 
         assert self._session is not None, "Session has not been created"
-        self._active_transaction = await self._session.begin_transaction()
+        timeout = get_config().transaction_timeout if timeout is None else timeout
+        self._active_transaction = await self._session.begin_transaction(
+            timeout=timeout
+        )
 
     @ensure_connection
     async def commit(self) -> Bookmarks:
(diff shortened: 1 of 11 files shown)


mariusconjeaud:
Oops, wrongly deleted, I messed up here...

PR #940, opened 2026-06-09
Fix unicode case-insensitive filters

profatsky:
# Fix unicode case-insensitive filter matching

## What changed

- Use `(?iu)` instead of `(?i)` for case-insensitive regex filters.
- Keep `iregex` patterns unescaped. Previously `iregex` could escape regex syntax such as `.*` because it shared the same template as `iexact`.

## Why

Unicode text did not match correctly with case-insensitive filters:

```python
UserNode(name="Алиса").save()
UserNode.nodes.filter(name__iexact="алиса").get_or_none() # => None
```

This affected `iexact`, `icontains`, `istartswith`, `iendswith`, and `iregex`.

## Tests

Added separate async and sync tests for Unicode case-insensitive filters.


deniskorbakov (approved):

mariusconjeaud:
Please pull from target branch rc/6.2.0 to have the tests fixed for the latest Neo4j versions

mariusconjeaud (approved):

profatsky:
Pull is done

PR #966, opened 2026-09-11
Fixes #965

### LINKED ISSUE #965: Hosted docs on readthedocs.io are stuck at 6.0.1 — every RTD build since Dec 2025 fails (missing typing_extensions)
### Summary

The documentation hosted at https://neomodel.readthedocs.io (both `/en/latest/` and `/en/stable/`) is stale: it still serves the **6.0.1** build from **Dec 29, 2025**. None of the changes from 6.1.0, 6.2.0, or the recent **7.0.0** release are visible. For example:

- `https://neomodel.readthedocs.io/en/latest/pydantic.html` (the new Pydantic bridge page) → 404
- The nav banner on both `latest` and `stable` reads "neomodel 6.0.1 documentation"
- Nothing on the hosted site documents 7.0.0's breaking change (`config.DATABASE_URL` no longer defaults to a local bolt URL), lazy class discovery, or the bulk operations

The doc sources under `doc/source/` on master are up to date — the problem is purely on the Read the Docs build side.

### Root cause

Every RTD build since March 2026 fails during the Sphinx config step:

```
Configuration error!
...
ModuleNotFoundError: No module named 'typing_extensions'
sphinx.errors.ConfigError: There is a programmable error in your configuration file:
```

(exact location varies by version — e.g. `neomodel/async_/match.py:7` in the 6.x-era builds, `neomodel/properties.py:11` in the 7.0.0 build)

`doc/source/conf.py` imports the package itself (`from neomodel import __author__, __package__`), and the package source has required `typing_extensions` (for `Self`) since ~6.1. But `doc/requirements.txt` only contains:

```
sphinx_copybutton
neo4j~=6.0.3
```

so `typing_extensions` is never installed into the build environment (Sphinx 9, used by RTD unpinned, no longer pulls it in transitively), and the build dies before rendering a single page.

### Evidence (RTD build history)

| Build | Date | Commit | Result |
|---|---|---|---|
| [#30847401](https://app.readthedocs.org/projects/neomodel/builds/30847401/) | 2025-12-29 | `7139fc2` (6.0.1) | ✅ last success — this is what's served today |
| [#31659435](https://app.readthedocs.org/projects/neomodel/builds/31659435/) | 2026-03 | `92e2059` | ❌ `ModuleNotFoundError: typing_extensions` |
| [#33265601](https://app.readthedocs.org/projects/neomodel/builds/33265601/) | 2026-06 | `7d0bde8` (6.2.0) | ❌ same error |
| [#34240627](https://app.readthedocs.org/projects/neomodel/builds/34240627/) | 2026-08-26 | `94e74f2` (7.0.0) | ❌ same error |

Side note: the builds *listing* page shows all of these with a "Built" badge, which makes the failures easy to miss — only the individual build pages / raw logs show the error.

### Suggested fix

Either:

1. Add `typing_extensions` to `doc/requirements.txt`, or (more robust)
2. Install the package itself in `.readthedocs.yaml` so docs builds always run against the real runtime dependencies:

```yaml
python:
  install:
    - requirements: doc/requirements.txt
    - method: pip
      path: .
```

Option 2 also prevents this class of failure for any future runtime dependency. Once fixed, rebuilding `latest`/`stable` would make the 7.0.0 docs — including the breaking `config.DATABASE_URL` change — visible to users.

aanastasiou:


aanastasiou:
@mariusconjeaud , any ideas?

image

aanastasiou:
Docs build from PR branch. The webhook seems to be fine. Possibly even ithout a refresh in the end. The message about not being able to install the webhook persists. Will check again later.

Will switch back to `master` once merged. In any case, I think it might be a good idea to migrate to the "app" if that is where things are going. Not the most urgent thing right now though.

aanastasiou:
The error is still being reported. 

Screenshot 2026-09-22 21 29 24

mariusconjeaud:
The webhook is set on the GitHub side, maybe there is something to do on the RTD side ? Is it trying to add it automatically, instead of picking up the one I created ?

PR #968, opened 2026-09-17
Fix async impersonation bug, remove print statements and write new tests for impersonation decorator 

### LINKED ISSUE #967: Issue with async impersonation transaction handling (Enterprise Edition)
## Expected Behaviour
When decorating an async neomodel transaction function with a handler for adb.impersonate(user="ExampleUser"), all queries executed by the wrapped function should run as the impersonated user.  

ie. 
```python
handler = await adb.impersonate(user="ExampleUser")

@handler
async def async_impersonate_example():
    results, _ = await adb.cypher_query("SHOW CURRENT USER YIELD user as CURRENT_USER")
    return results[0][0]

user = await async_impersonate_example() # Expect: user = "ExampleUser"
```

I can confirm that this expected behaviour is met in the synchronous  circumstance. 

## Actual Behaviour
Queries executed with the decorator run as the user specified in the db configuration

ie. 
```python
from neomodel import adb
await adb.set_connection(url="bolt://neo4j:foobarbaz@localhost:7687")

handler = await adb.impersonate(user="ExampleUser")

@handler
async def async_impersonate_example():
    results, _ = await adb.cypher_query("SHOW CURRENT USER YIELD user as CURRENT_USER")
    return results[0][0]

user = await async_impersonate_example() # Actual: user = "neo4j"
```
## How to Reproduce the Problem
1. Start a Neo4j EE 
2. Create an impersonatable user 
```
CREATE OR REPLACE USER ExampleUser SET PASSWORD 'example' CHANGE NOT REQUIRED
GRANT ROLE publisher TO ExampleUser
GRANT IMPERSONATE (ExampleUser) ON DBMS TO admin
```
3. Run the script below
```python
import asyncio
from neomodel import adb

async def current_user() -> str:
    results, _ = await adb.cypher_query("SHOW CURRENT USER YIELD user as CURRENT_USER")
    return results[0][0]


async def main() -> None:
    IMPERSONATED_USER="ExampleUser"
    await adb.set_connection(url="bolt://neo4j:foobarbaz@localhost:7687")

    # Works as intended
    with await adb.impersonate(user=IMPERSONATED_USER):
        print(await current_user())  # Impersonation was applied | Actual: ExampleUser
    
    handler = await adb.impersonate(user=IMPERSONATED_USER)

    @handler # This is BROKEN
    async def decorated() -> str:
        return await current_user()

    print(await decorated())  # Impersonation was NOT applied | Actual: neo4j 

    await adb.close_connection()


asyncio.run(main())
```
```console
$ python above_script.py 
ExampleUser

Exception type: None

Exception value: None

Traceback: None

Exception type: None

Exception value: None

Traceback: None
neo4j

```
You will also see that some debugging print statements are still run with this use

### Versions
  - OS: macOS (darwin)
  - Library: neomodel 7.0.0, neo4j Python driver 6.1.0
  - Neo4j: 2026.03-enterprise

greengori11a:
* Remove print statements
* Bug Fix
* Create async and sync test for decorator impersonation

Fixes #967


aanastasiou:
Hi @greengori11a , thanks for this, is this the final PR?

greengori11a:
@aanastasiou - yes it is, i believe the bug is fixed and the tests pass. is there anything you would like me to change or do differently?

aanastasiou:
@greengori11a Not at the moment, just wanted to make sure the final piece of work has been included.

